PR #3, opened 2021-11-07
Add support for manifest list/image index manifest

tofay:
This PR is an initial attempt to add support for manifest lists. (I wasn't sure whether to submit to here or krustlet/krustlet, so just picked here)

The approach I've gone for is:
- add OciImageIndexManifest struct representing an image manifest
- rename existing OciManifest to OciImageManifest
- create OciManifest enum wrapping the Image and Image Index manifest
- users can specify how a client behaves when it receives an image index manifest via the platform_resolver field
  - by default the client attempts to select an image from the image index manifest that matches the running platform
- the exception is the pull_manifest function - if someone is just pulling a manifest then they should get the manifest as it is in the registry
   - I've added a pull_image_manifest function which does attempt platform resolution

This approach gives users the ability to control how the client deals with image index manifests, without forcing users to deal with manifests on every image pull.

Obviously that's breaking, particularly the OciManifest struct- > enum switch. Though if this crate is to support manifest lists then I think OciManifest shouldn't be used to refer to image manifests specifically.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -5,8 +5,9 @@
 
 use crate::errors::*;
 use crate::manifest::{
-    OciDescriptor, OciManifest, Versioned, IMAGE_LAYER_GZIP_MEDIA_TYPE, IMAGE_LAYER_MEDIA_TYPE,
-    IMAGE_MANIFEST_MEDIA_TYPE,
+    ImageIndexEntry, OciDescriptor, OciImageManifest, OciManifest, Versioned,
+    IMAGE_LAYER_GZIP_MEDIA_TYPE, IMAGE_LAYER_MEDIA_TYPE, IMAGE_MANIFEST_LIST_MEDIA_TYPE,
+    IMAGE_MANIFEST_MEDIA_TYPE, OCI_IMAGE_INDEX_MEDIA_TYPE, OCI_IMAGE_MEDIA_TYPE,
 };
 use crate::secrets::RegistryAuth;
 use crate::secrets::*;
@@ -27,10 +28,10 @@ use tracing::{debug, trace, warn};
 use www_authenticate::{Challenge, ChallengeFields, RawChallenge, WwwAuthenticate};
 
 const MIME_TYPES_DISTRIBUTION_MANIFEST: &[&str] = &[
-    "application/vnd.docker.distribution.manifest.v2+json",
-    // TODO: support manifest lists?
-    // "application/vnd.docker.distribution.manifest.list.v2+json",
-    "application/vnd.oci.image.manifest.v1+json",
+    IMAGE_MANIFEST_MEDIA_TYPE,
+    IMAGE_MANIFEST_LIST_MEDIA_TYPE,
+    OCI_IMAGE_MEDIA_TYPE,
+    OCI_IMAGE_INDEX_MEDIA_TYPE,
 ];
 
 /// The data for an image or module.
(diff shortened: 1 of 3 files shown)


thomastaylor312:
@tofay This is the place to do it! It might take a sec for some of us to get to reviewing this due to some other time constraints.

bacongobbler:
Sorry! I forgot to subscribe to new notifications and didn't see this. Right now just working on getting the next release out - once that's done I'll review this. Thanks for being patient!

bacongobbler:
okay, v0.8.0 is out of the way. Would you mind rebasing?

tofay:
Done. 

I noticed that Cargo.lock is incompatible in main - I'm happy to fix separately by removing Cargo.lock or updating CI to verify that it's compatible.

bacongobbler:
Sorry not quite sure what you mean. When you say "incompatible in main", what is incompatible? Do you have an error or something to demonstrate? I don't see anything after running `cargo build` on main. I'm not sure what I should be looking for.

tofay:
Sorry for not being clear. `cargo build --locked` fails because the oci-distribution version in Cargo.lock is 0.7.0.

bacongobbler:
Works for me on a fresh clone:

```
bacongobbler@broodlord ~/code/krustlet ><> git status
On branch main
Your branch is up to date with 'origin/main'.

nothing to commit, working tree clean
bacongobbler@broodlord ~/code/krustlet ><> cargo build --locked
    Finished dev [unoptimized + debuginfo] target(s) in 0.11s
```

Though I do see oci-distribution at v0.7.0 in Cargo.lock.

Any suggestions how to fix that?

bacongobbler:
I just did some research RE: your earlier question on whether we should remove it entirely.

Reading https://github.com/rust-lang/cargo/issues/315 suggests that we should remove it since this is a library.

I'll go ahead and do that in a follow-up PR.

bacongobbler:
d'oh. I just realized I was running all of these commands from the krustlet repository and not oci-distribution, which may explain why I'm not seeing the same things as you are. Clearly I need more coffee this morning. ☕

In any case I'll go ahead with the suggestion in my last comment - that should hopefully fix things up for you

bacongobbler:
done in #4. Try rebasing again :)

src/client.rs
```diff
@@ -486,6 +512,55 @@ impl Client {
         self._pull_manifest(image).await
     }
 
+    /// Pull an image manifest from the remote OCI Distribution service.
+    ///
+    /// If the connection has already gone through authentication, this will
+    /// use the bearer token. Otherwise, this will attempt an anonymous pull.
```
bacongobbler, line 518: Why do we need to pull this out into a separate function? Why not wrap `auth` as an `Option<&RegistryAuth>` instead?

tofay: I was using the same pattern as _pull_manifest. I assume that pattern is there so that the external API mandates a RegistryAuth, but internally that may not be required (as it's already auth has already been done).

arronwy:
@bacongobbler @tofay @thomastaylor312  any updates for this PR? Thought manifest list support is optional, it is widely used for OCI images to support different platforms.

src/client.rs
flavio: Is this conversation still relevant? I think we can mark that as resolved

flavio (changes requested):
This looks good to me.

There are super minor comment I left, plus it needs a rebase

src/client.rs
tofay: marking as resolved based on my previous comment

tofay:
I've rebased (including undoing https://github.com/krustlet/oci-distribution/pull/19 ) and addressed the comments.
Thanks for the review.

src/client.rs
```diff
@@ -991,6 +1061,86 @@ pub struct ClientConfig {
...
+            protocol: ClientProtocol::default(),
+            accept_invalid_hostnames: false,
+            accept_invalid_certificates: false,
+            extra_root_certificates: Vec::new(),
+            platform_resolver: Some(Box::new(current_platform_resolver)),
+        }
+    }
+}
+
+type PlatformResolverFn = dyn Fn(&[ImageIndexEntry]) -> Option<String>;
```
bacongobbler, line 1087: Instead of this all being in the client package, perhaps this would be useful in a separate module. Doesn't have to be done in this PR, but just thought this would be useful outside of the context of the OCI client.

bacongobbler (approved):
LGTM

PR #6, opened 2021-11-25
Fix interacting with GitHub Container Registry

flavio:
The GitHub container registry (ghcr.io) returns a bearer token that doesn't have an explicit expiration time.
    
Prior to this change, because of this anomaly, the token was not added into the JWT token cache. That caused all the request done against ghcr.io to be unauthenticated. That broke any kind of interaction with ghcr, even with public images.
    
This code handles the lack of an expiration claim by setting a 60 second expiration window.
According to [the official documentation](https://docs.docker.com/registry/spec/auth/token/#requesting-a-token), this behaviour is fine:
    
> The duration in seconds since the token was issued that it will remain valid.
> When omitted, this defaults to 60 seconds. For compatibility with older clients,
> a token should never be returned with less than 60 seconds to live.

### DIFF
--- a/src/token_cache.rs
+++ b/src/token_cache.rs
@@ -1,6 +1,8 @@
 use crate::reference::Reference;
 use serde::Deserialize;
 use std::collections::BTreeMap;
+use std::fmt;
+use std::time::{SystemTime, UNIX_EPOCH};
 use tracing::{debug, warn};
 
 /// A token granted during the OAuth2-like workflow for OCI registries.
@@ -12,6 +14,22 @@ pub(crate) enum RegistryToken {
     AccessToken { access_token: String },
 }
 
+impl fmt::Debug for RegistryToken {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let redacted = String::from("<redacted>");
+        match self {
+            RegistryToken::Token { .. } => {
+                f.debug_struct("Token").field("token", &redacted).finish()
+            }
+            RegistryToken::AccessToken { .. } => f
+                .debug_struct("AccessToken")
+                .field("access_token", &redacted)
+                .finish(),
+        }
+    }
+}
+
+#[derive(Debug)]
 pub(crate) enum RegistryTokenType {
     Bearer(RegistryToken),
     Basic(String, String),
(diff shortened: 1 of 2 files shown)


src/token_cache.rs
```diff
@@ -1,17 +1,19 @@
...
-#[derive(Deserialize, Clone)]
+#[derive(Deserialize, Debug, Clone)]
 #[serde(untagged)]
 #[serde(rename_all = "snake_case")]
 pub(crate) enum RegistryToken {
     Token { token: String },
     AccessToken { access_token: String },
 }
 
+#[derive(Debug)]
```
thomastaylor312, line 16: I worry about deriving debug here as it could leak sensitive data when used in a `println!`. If we need a debug, we should manually implement it so as to mask any sensitive data

src/token_cache.rs
```diff
@@ -66,12 +68,26 @@ impl TokenCache {
...
                         jwt::claims::Claims,
                         jwt::token::Unverified,
-                    >::parse_unverified(token_str)
+                    >::parse_unverified(&token_str)
                     {
                         Ok(token) => token.claims().registered.expiration.unwrap_or(u64::MAX),
                         Err(error) => {
-                            warn!(?error, "Invalid bearer token");
-                            return;
+                            // the token doesn't have a claim that states a
```
thomastaylor312, line 75: This logic looks good but assumes that the error returned is due to the missing expiry. This `Err` block should contain a guard using `matches!` or whatever else is needed to make sure we are handling only the error case of a missing expiry. Other error cases should continue with the warning like what existed before

thomastaylor312 (changes requested):

src/token_cache.rs
flavio: Good point, I've fixed that

src/token_cache.rs
flavio: Good point, I've fixed that now

flavio:
@thomastaylor312 I've made two commits:
  * https://github.com/krustlet/oci-distribution/pull/6/commits/bf716dbf67b82831e9562eb53382bc4b3a6c26d5 fixes the issues you reported during your review
  * https://github.com/krustlet/oci-distribution/pull/6/commits/4a245ff2a2ee7f89ff10807f75285396a7da4349 adds a new basic unit test to ensure GHCR works fine

I need feedback about the unit test I've added. I picked an OCI image stored inside of a ghcr repository that I own, but I feel like we should have something owned/controller by oci-distribution. What do you think?

As a last note, once everything is looking good I can squash the commits to reduce the pollution on the branch

thomastaylor312 (approved):
This all looks great and good call on the additional test. I definitely agree with you that we should use an oci-distribution maintained image. If you can't find one, I think the kubewarden would work fine.

Approving this either way. No need for another re-review if you switch the image!

flavio:
> If you can't find one, I think the kubewarden would work fine.

We could push a simple wasm file, like one that is already being used by the unit tests, under the `krustlet` organization on GitHub

I think I might have the rights to do that if I submit the wasm file to `ghcr.io/krustlet/oci-distribution/something-wasm:0.0.1`. Would you be fine with that?

thomastaylor312:
Actually that is a great idea! Let's do that

flavio:
@thomastaylor312 I don't have the rights to do this operation.

Can you do that on my behalf, please?

```console
wasm-to-oci pull webassembly.azurecr.io/hello-wasm:v1
wasm-to-oci push module.wasm ghcr.io/krustlet/oci-distribution/hello-wasm:v1
```

I'm pretty sure you've already `wasm-to-oci` installed on your machine ;)

thomastaylor312:
Ok, done!

flavio:
> Ok, done!

This artifact is still marked as private on the GHCR:

```console
wasm-to-oci pull ghcr.io/krustlet/oci-distribution/hello-wasm:v1
Error: unexpected status code [manifests v1]: 401 Unauthorized
```

Moreover, if you go [here](https://github.com/krustlet/oci-distribution), you will notice there's nothing shown inside of the "Packages" section (which is below "Releases").

You can fix that using the GH web UI. AFAIK you have to go to the packages tab of the krustlet organization, select the repository and make it public

thomastaylor312:
Sorry @flavio! I haven't really used ghcr before. Should be public now

flavio:
> Sorry @flavio! I haven't really used ghcr before. Should be public now

No worries!

Everything is working as expected now. Do you want me to rebase the commits a little, just to keep the history more clean?

flavio:
The test failing is not related with my changes. It's probably worth a dedicated issue

thomastaylor312:
@flavio Yeah, that is a separate thing. Feel free to merge when ready. You can rebase if you'd like or just do a squash and merge with the button

PR #8, opened 2021-12-02
Make accept_invalid_hostnames conditional

ereslibre:
This settings depends on the SSL backend, and is only applicable if
the openssl native backend is used.


thomastaylor312 (approved):

thomastaylor312:
Thanks @ereslibre!

Issue #7, opened 2021-11-25
Move `src/www_authenticate` into a sub-crate


flavio:
The directory `src/www_authenticate` contains a copy of the [www-authenticate](https://crates.io/crates/www-authenticate) crate. I realize this is an approach that was taken when `oci-distribution` was a sub-crate of `krustlet`.

I propose to turn this module into a sub-crate, so that `cargo build|check|...` stops producing tons of warning messages, one per public method of the `www-authenticate` module that `oci-distribution` doesn't use.

A quick example:

```console
warning: associated function is never used: `get_mut`
   --> src/www_authenticate/mod.rs:253:16
    |
253 |         pub fn get_mut(&mut self, k: &str) -> Option<&mut String> {
    |                ^^^^^^^

warning: associated function is never used: `insert_quoting`
   --> src/www_authenticate/mod.rs:265:16
    |
265 |         pub fn insert_quoting(&mut self, k: String, v: String) -> Option<String> {
    |                ^^^^^^^^^^^^^^

warning: `oci-distribution` (lib) generated 14 warnings
    Finished dev [unoptimized + debuginfo] target(s) in 1.03s
```

In theory, moving everything into a sub-crate should allow us to get rid of these warnings

flavio:
@kflansburg, @thomastaylor312, @bacongobbler: what do you think about this proposal? If you are fine with that I can give it a shot and see if I can make the warnings disappear

thomastaylor312:
I am definitely in favor of making this a subcrate

flavio:
According to our README, we have this code copied into the oci-distribution repository because:

> This code was copied from https://github.com/KeenS/www-authenticate, which is unlicensed.

I double checked, and now the crate has a MIT license. The license is stated explicitly inside of its [Cargo.toml](https://github.com/KeenS/www-authenticate/blob/612fbd949d137f2bd7b41b3a072ab3932a813bd6/Cargo.toml#L5) file, but there's no `LICENSE` file inside of the git repository.

Should I go ahead with this issue or should we consume the crate from `crates.io`?

I just want to understand the whole issue before doing this change :pray:

PR #10, opened 2021-12-14
move www-authenticate to a sub-crate

flavio:
Move the `www-authenticate` to a dedicated sub-crate.

This reduces the number of warnings generated at build time.

On top of that, other small fixes have been done to address the clippy warnings that became visible.

Fixes issue #7


thomastaylor312:
@flavio So I just realized I may have screwed up. I don’t know if we can publish the crate anymore if we make it a subcrate. Can you try a `cargo publish --dry-run` on the main oci-distribution crate and see if it gives you a warning about unpublished crates? If it is, I apologize for sending you down this rabbit hole

flavio:
ah! you're right, this is not going to fly:

```console
$ cargo publish --dry-run 
    Updating crates.io index
error: all dependencies must have a version specified when publishing.
dependency `www-authenticate` does not specify a version
Note: The published dependency will use the version from crates.io,
```

What about:
  * fork the `www-authenticate` repository under the krustlet organization
  * add your patches inside of it
  * tag a new version of it
  * update `oci-distribution` crate to get the dependency from our git fork, using the tag we just created

This would also have the advantage of making more evident the changes we've done to the original `www-authenticate` crate.

Could that work?

flavio:
I fear this would not work either according to the rust docs (thanks @ereslibre for pointing that out): https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations

I fear the only solution would be to create a fork of www-authenticate with our changes, and then publish that with a different name on crates.io

thomastaylor312:
Yeah we can't publish with the same name. I think we just keep it as is for now and then update it when the maintainer eventually cuts another version

bacongobbler (approved):

bacongobbler:
I agree with @thomastaylor312's thoughts. The only reason the crate was pulled in here was because the project [appears to be unmaintained](https://github.com/KeenS/www-authenticate/issues/4#issuecomment-832889092) and we haven't seen a release since 2018.

thomastaylor312:
@bacongobbler @flavio  Should we go ahead and close this PR then?

flavio:
Yep, closing. I'll file a PR against the original repo and see how the maintainer reacts to it

PR #11, opened 2021-12-22
Remove in-tree version of www-authenticate

flavio:
The in-tree version of www-authenticate can be removed because the library maintainer has released a new version of the crate that addresses all our needs.

Fixes https://github.com/krustlet/oci-distribution/issues/7


thomastaylor312 (approved):
Awesome! Thanks for driving stuff forward with the www-authenticate crate!

flavio:
Sorry about the delay, I was on vacation :)

Can I go ahead and merge it? The CI is failing because of an unrelated issue (see https://github.com/krustlet/oci-distribution/issues/13)

thomastaylor312:
Yeah, I just need to set aside some time to look at that

PR #19, opened 2022-01-24
Remove manifest list from MIME Accept header

kflansburg:
We pass a list of accepted MIME types when pulling manifests, but then we immediately reject anything other than `application/vnd.docker.distribution.manifest.v2+json` when validating the response on:
https://github.com/krustlet/oci-distribution/blob/0e5373e99c910ffd1261ace7e1b732036dc7784a/src/client.rs#L548

This change removes the list type which we are incapable of parsing. As a result, registries which can serve manifest lists (like DockerHub) will return a single manifest instead (presumably for `amd64`?). If we have a strong opinion in the future about which manifest to use from the list, we can implement proper parsing for these manifest lists.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -28,7 +28,8 @@ use www_authenticate::{Challenge, ChallengeFields, RawChallenge, WwwAuthenticate
 
 const MIME_TYPES_DISTRIBUTION_MANIFEST: &[&str] = &[
     "application/vnd.docker.distribution.manifest.v2+json",
-    "application/vnd.docker.distribution.manifest.list.v2+json",
+    // TODO: support manifest lists?
+    // "application/vnd.docker.distribution.manifest.list.v2+json",
     "application/vnd.oci.image.manifest.v1+json",
 ];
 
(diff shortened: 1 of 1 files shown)


thomastaylor312:
I am not confident enough with my OCI knowledge to know if this change is ok. @flavio do you have any insight here?

rickrain:
Would really like to see this ( or #3 ) merged in soon. This one is perfect for now IMO and wouldn't break anything since it (`manifest.list`) was not supported anyway.

flavio (approved):
Sorry for the delay, I think it's safe to merge this PR. I'll also take a closer look at #3

flavio:
@kflansburg could you please add your signature to commit 94a3993215aae7658d36498cd3f9cb73c61d3502? This is required to make the DCO check green

kflansburg:
> @kflansburg could you please add your signature to commit [94a3993](https://github.com/krustlet/oci-distribution/commit/94a3993215aae7658d36498cd3f9cb73c61d3502)? This is required to make the DCO check green

Done

PR #21, opened 2022-02-01
Change the API of Client::push

flavio:
Previous to this commit, `Client::push` would return a `Result<String>` object. The embedded String would be the URL of the image just pushed to the registry.

There are however more URLs that are returned by an OCI registry upon a successful push: image URL, config URL and manifest URL. 
These URLs cab be useful too, hence the code has been changed so that all of them are returned back.

The manifest URL for example can be pretty useful. The URL contains the manifest digest of the image, which can be used to pull the contents in an immutable way.

Having the manifest digest at hand is useful also when a freshly pushed OCI image/artifact has to be signed with something like Sigstore.
In that case, it's a best practice to sign the content by its immutable address, rather than by its tag.

Having the `Client::push` API return the manifest digest allows safer signing of contents that prevents both race conditions and malicious actors (including the registry).

For more details check this detailed blog post about Sigstore: https://blog.chainguard.dev/zero-friction-keyless-signing-with-github-actions/

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -64,6 +64,17 @@ impl ImageData {
     }
 }
 
+/// The data returned by an OCI registry after a successful push
+/// operation is completed
+pub struct PushResponse {
+    /// Pullable url for the image
+    pub image_url: String,
+    /// Pullable url for the config
+    pub config_url: String,
+    /// Pullable url for the manifest
+    pub manifest_url: String,
+}
+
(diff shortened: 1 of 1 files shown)


thomastaylor312 (approved):
This seems like a great change!

thomastaylor312:
Labeling as breaking so we know to call this change out in the next release notes

PR #22, opened 2022-02-03
Fix various with image pushes.

tofay:
I've been trying to use this crate to pull images from one registry and push them to another - I've fixed a few issues I encountered, and added a test for roundtrip pull/pushing an image with multiple layers.

The existing push doesn't work for images with multiple layers because the code conflates chunks and layers. e.g it starts a push session for ImageData, then proceeds to push all the layers as chunks in a single blob.
I've updated this to have a push session per layer, per the spec.

The push response for `Client::push` similarly doesn't return an image_url, as that
doesn't make sense for images with multiple layers.

The ImageData struct no longer produces a digest from the contained
layers as digests are applicable to content blobs, or manifests.

If the manifest being pushed has a media type, then use that instead of the OCI V1 media type. (If the manifest is not the OCI v1 media type then the registry will reject the manifest push.)

Manifests are canonicalized to canonical json per
https://github.com/opencontainers/image-spec/blob/main/considerations.md?plain=1#L18.
I've also updated the manifest structs to not serialize unspecified optional fields as null. That complies with the json schemas in the image-spec repo. The spec wasn't clear on the behaviour here, so I raised https://github.com/opencontainers/image-spec/issues/894. I probably should have done that in a separate PR - happy to drop it from this and submit in a new one if that's contravertial.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,4 +51,4 @@ olpc-cjson = "0.1.1"
 [dev-dependencies]
 rstest = "0.11"
 hmac = "0.11"
-env_logger = "0.9.0"
+tracing-subscriber = "0.3.7"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,9 @@ tracing = {version = "0.1", features = ['log']}
 unicase = "1"
 url = "1"
 www-authenticate = "0.4.0"
+olpc-cjson = "0.1.1"
 
 [dev-dependencies]
 rstest = "0.11"
 hmac = "0.11"
+env_logger = "0.9.0"
--- a/src/client.rs
+++ b/src/client.rs
@@ -1319,10 +1319,6 @@ mod test {
     const GHCR_IO_IMAGE: &str = "ghcr.io/krustlet/oci-distribution/hello-wasm:v1";
     const DOCKER_IO_IMAGE: &str = "docker.io/library/hello-world@sha256:37a0b92b08d4919615c3ee023f7ddb068d12b8387475d64c622ac30f45c29c51";
 
-    fn init() {
-        let _ = env_logger::builder().is_test(true).try_init();
-    }
-
     #[test]
     fn test_apply_accept() -> Result<(), anyhow::Error> {
         assert_eq!(
@@ -2003,7 +1999,7 @@ mod test {
     #[ignore]
     /// Requires local registry resolveable at `oci.registry.local`
     async fn test_image_roundtrip() {
-        init();
+        tracing_subscriber::fmt::init();
         let mut c = Client::new(ClientConfig {
             protocol: ClientProtocol::HttpsExcept(vec!["oci.registry.local".to_string()]),
             ..Default::default()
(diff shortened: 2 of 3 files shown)


src/client.rs
```diff
@@ -240,53 +307,55 @@ impl Client {
         &mut self,
         image_ref: &Reference,
         image_data: &ImageData,
-        config_data: &[u8],
```
tofay, line 243: This API feels cumbersome for users pushing something that they haven't just pulled, as in that case they need to create an ImageData. 

perhaps it would be better as the following? 
```
pub async fn push(
        &mut self,
        image_ref: &Reference,
        layers: Vec<ImageLayer>,
        config: Config,
        auth: &RegistryAuth,
        manifest: Option<OciImageManifest>)
```

Cargo.toml
```diff
@@ -46,7 +46,9 @@ tracing = {version = "0.1", features = ['log']}
 unicase = "1"
 url = "1"
 www-authenticate = "0.4.0"
+olpc-cjson = "0.1.1"
 
 [dev-dependencies]
 rstest = "0.11"
 hmac = "0.11"
+env_logger = "0.9.0"
```
shanesveller, line 54: This crate already holds a full dependency on `tracing` with `log` feature enabled, please lean on that and/or take a dev-dependency on `tracing-subscriber` instead:

https://docs.rs/tracing/latest/tracing/#for-log-users
https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.Layer.html#method.with_test_writer

tofay: Thanks, done.

src/client.rs
flavio: Yes, I think it would provide a better UX

src/client.rs
```diff
@@ -2085,4 +2105,35 @@ mod test {
...
+    #[tokio::test]
+    #[ignore]
+    async fn test_roundtrip_multiple_layers() {
+        tracing_subscriber::fmt::init();
+        let mut c = Client::new(ClientConfig {
+            protocol: ClientProtocol::HttpsExcept(vec!["oci.registry.local".to_string()]),
+            ..Default::default()
+        });
+        let src_image = Reference::try_from("registry:2.7.1").expect("failed to parse reference");
+        let dest_image = Reference::try_from("oci.registry.local/python:test")
```
flavio, line 2118: nitpicking, I think the target destination should be different

src/client.rs
```diff
@@ -2085,4 +2105,35 @@ mod test {
             .unwrap();
         assert_eq!(manifest.config.media_type, manifest::WASM_CONFIG_MEDIA_TYPE);
     }
+
+    #[tokio::test]
+    #[ignore]
```
flavio, line 2110: Unrelated with this PR, I think we should have a way to run this unit test (and the other ignored ones) in automated fashion.

flavio (commented):
Looks good to me, I left some minor comments

src/client.rs
tofay: Agree - one of the ignored tests is failing prior to this.

I'll take a look at using testcontainer-rs for the ignored tests separately.

src/client.rs
tofay: ah yes, I've updated

src/client.rs
tofay: I raised https://github.com/krustlet/oci-distribution/pull/24 to address this.

flavio (approved):
Thanks for having addressed the feedback

flavio:
@cardil :+1: I personally would like to go through a bunch of open PR, so that we break the API fewer times :)

PR #24, opened 2022-02-08
replace ignored tests with testcontainers

tofay:
This PR updates the tests that require pushing to a local registry to instead launch a registry container using testcontainers.


tofay:
Leaving as draft until https://github.com/krustlet/oci-distribution/pull/22 is in, as one of the ignored tests currently fails.

flavio (commented):
That's brilliant. I didn't know about the *testcontainers* project.

I love this PR, thanks a lot!

flavio (approved):
LGTM

Cargo.toml
```diff
@@ -52,3 +54,4 @@ olpc-cjson = "0.1.1"
 rstest = "0.11"
 hmac = "0.11"
 tracing-subscriber = "0.3.7"
+testcontainers = "0.12.0"
```
bacongobbler, line 57: perhaps this should be under `dev-dependencies`?

https://doc.rust-lang.org/rust-by-example/testing/dev_dependencies.html

bacongobbler (changes requested):
Looks great! Just had one suggested change.

Cargo.toml
bacongobbler: Ah... It already is. Github didn't expand the list far enough. LGTM!

bacongobbler (approved):

PR #25, opened 2022-02-09
fix: Allow build with feature `rustls-tls`

viccuad:
This change maintains the previous behaviour on Windows, but I'mt not
sure that is correct to be fair, and I suspect we only need:

`#[cfg(feature = "native-tls")]`

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1093,7 +1093,7 @@ impl Default for ClientConfig {
     fn default() -> Self {
         Self {
             protocol: ClientProtocol::default(),
-            #[cfg(not(target_os = "windows"))]
+            #[cfg(feature = "native-tls")]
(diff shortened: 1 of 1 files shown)


flavio:
@viccuad I agree with you, I think the macro should be changed to be a simple:

```rust
#[cfg(feature = "native-tls")]
```

That's because https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.danger_accept_invalid_hostnames is available only when native-tls is being used, regardless of windows...

viccuad:
Indeed, thanks for the review! Force-pushed amended commit with `#[cfg(feature = "native-tls")]` .

flavio (approved):
LGTM

PR #26, opened 2022-02-10
Added ability to push manifest lists & added ease-of-use impls

jsoverson:
This crate didn't seem to allow pushing of manifest lists and the private authentication calls made it difficult to work in without a change.

The changes beyond adding `Client::push_manifest_list` alone come from working on the crate. They're mostly implementations of `From` or `Display` to make debugging easier.

- Added public function Client::push_manifest_list
- Added public function OciManifest::content_type
- Added Display impl for OciManifest and variant types
- Added From impl for OciManifest and variant types
- Added IMAGE_DOCKER_LAYER_TAR_MEDIA_TYPE type
- Skipped serializing for Option::None fields in manifest types

### DIFF
--- a/src/manifest.rs
+++ b/src/manifest.rs
@@ -1,6 +1,8 @@
 //! OCI Manifest
 use std::collections::HashMap;
 
+use serde::Serialize;
+
 /// The mediatype for WASM layers.
 pub const WASM_LAYER_MEDIA_TYPE: &str = "application/vnd.wasm.content.layer.v1+wasm";
 /// The mediatype for a WASM image config.
@@ -22,6 +24,8 @@ pub const IMAGE_DOCKER_CONFIG_MEDIA_TYPE: &str = "application/vnd.docker.contain
 pub const IMAGE_LAYER_MEDIA_TYPE: &str = "application/vnd.oci.image.layer.v1.tar";
 /// The mediatype for a layer that is gzipped.
 pub const IMAGE_LAYER_GZIP_MEDIA_TYPE: &str = "application/vnd.oci.image.layer.v1.tar+gzip";
+/// The mediatype that Docker uses for a layer that is tarred.
+pub const IMAGE_DOCKER_LAYER_TAR_MEDIA_TYPE: &str = "application/vnd.docker.image.rootfs.diff.tar";
 /// The mediatype that Docker uses for a layer that is gzipped.
 pub const IMAGE_DOCKER_LAYER_GZIP_MEDIA_TYPE: &str =
     "application/vnd.docker.image.rootfs.diff.tar.gzip";
@@ -44,6 +48,16 @@ pub enum OciManifest {
     ImageIndex(OciImageIndex),
 }
 
+impl OciManifest {
+    /// Returns the appropriate content-type for each variant.
+    pub fn content_type(&self) -> &str {
+        match self {
+            OciManifest::Image(_) => OCI_IMAGE_MEDIA_TYPE,
+            OciManifest::ImageIndex(_) => IMAGE_MANIFEST_LIST_MEDIA_TYPE,
+        }
+    }
+}
+
 /// The OCI image manifest describes an OCI image.
 ///
 /// It is part of the OCI specification, and is defined here:
@@ -62,6 +76,7 @@ pub struct OciImageManifest {
     ///
     /// It is reserved for compatibility, but the specification does not seem
     /// to recommend setting it.
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub media_type: Option<String>,
 
     /// The image configuration.
@@ -80,6 +95,7 @@ pub struct OciImageManifest {
     /// The specification says "If there are no annotations then this property
     /// MUST either be absent or be an empty map."
     /// TO accomodate either, this is optional.
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub annotations: Option<HashMap<String, String>>,
 }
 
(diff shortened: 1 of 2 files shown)


src/manifest.rs
```diff
@@ -44,6 +48,16 @@ pub enum OciManifest {
     ImageIndex(OciImageIndex),
 }
 
+impl OciManifest {
+    /// Returns the appropriate content-type for each variant.
+    pub fn content_type(&self) -> &str {
+        match self {
+            OciManifest::Image(_) => OCI_IMAGE_MEDIA_TYPE,
```
thomastaylor312, line 55: Nit: would it be better to return this as an actual `HeaderValue` rather than having to do the `parse().unwrap()` outside of this function?

src/manifest.rs
```diff
@@ -95,6 +111,42 @@ impl Default for OciImageManifest {
...
+}
+
+fn write_json(obj: &impl Serialize, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    write!(
+        f,
+        "{}",
+        serde_json::to_string(obj).map_err(|_| std::fmt::Error)?
+    )
+}
+impl std::fmt::Display for OciManifest {
```
thomastaylor312, line 132: Serializing JSON seems pretty heavy for using in a `Display` implementation, particularly `Display` is used in a lot of logging contexts which will incur a decent performance penalty each time it is logged. Perhaps this should be our custom debug implementation or we should drop it.

@flavio any thoughts here (since I think you use this library even more than I do)?

src/manifest.rs
```diff
@@ -95,6 +111,42 @@ impl Default for OciImageManifest {
...
+    }
+}
+impl From<OciImageManifest> for OciManifest {
+    fn from(m: OciImageManifest) -> Self {
+        Self::Image(m)
+    }
+}
+
+fn write_json(obj: &impl Serialize, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    write!(
```
thomastaylor312, line 126: If we do stick with using JSON for the `Display` implementation, I'd recommend we use `serde_json::to_writer` instead to make this as efficient as possible (it skips allocating a string and then writing it)

thomastaylor312 (changes requested):
One nit and one item of discussion. Overall this looks really great!

src/manifest.rs
jsoverson: I opted to keep it easily comparable to the content-type constants, but it was a coin-toss decision.

It would be nice to have an enum for the content-types and represent this as such, but that's a different topic.

src/manifest.rs
jsoverson: I understand the concern. Since JSON is how most people interact with manifests, I figured having that as the `Display` impl seemed appropriate. Debug output is still available for those who want that usage.

src/manifest.rs
jsoverson: Good point, I'll wait for the decision above and submit a change if JSON stays in.

src/manifest.rs
bacongobbler: [According to the docs](https://doc.rust-lang.org/std/fmt/trait.Display.html), `Display` is meant for user-facing output so I don't see an issue on preformance if we render the object as JSON.

That being said, I do have concerns over the ergonomics of this API. What if the caller wanted to display the object as YAML? What if the user wanted it to be in a "prettified" format? I could see others asking for an `output_format` field on the `OciManifest` to modify `Display`'s output... I don't think that's ideal.

I think we should follow the "format" defined in https://doc.rust-lang.org/rust-by-example/hello/print/print_display.html. In other words:

```
 fmt::Display for OciManifest {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        // display as "image: {}, manifest: {}"
        // ....
    }
}
```

flavio: @thomastaylor312  I rely a lot on `tracing` to handle my log messages. With `tracing` I can decided how an object is going to be handled. You could use the `Display`, the `Debug` or just do something ad-hoc. Hence, performance hits inside of logs are not something concerning me.

@bacongobbler I agree with @jsoverson, being the manifests JSON objects, it felt natural to have `Display` return a prettified JSON.

Having said that, I would be in favor of providing a more "standard" behaviour, like the one described above by @bacongobbler.

Users can then rely on `serde` or other libraries to pretty-print the manifest in their favourite way

thomastaylor312: Ok, so to summarize, I think we use the more standard way of implementing `Display` and leave it up to consumers of the crate to print how they want. If we find that most users are copy/pasting the same code in the future, we can always expose that as part of the crate

flavio: that sounds good to me @thomastaylor312

src/manifest.rs
flavio: I looked into doing that, unfortunately `Formatter` is implementing the [`std::fmt::Write`](https://doc.rust-lang.org/std/fmt/trait.Write.html) trait, not the [`std::io::Write`](https://doc.rust-lang.org/nightly/std/io/trait.Write.html) one which is required by `serde_json::to_writer`.

We have to leave the code as it is

src/manifest.rs
flavio: This PR has been pending since a long time, I would like to get this merged so that we can tag a new release of oci-distribution.

I've tried to implement the change requested, but I realized this would be pretty painful: we do have a lots of attributes inside of these Structs and the other Structs that belong to them.
Some of these attributes are also `Optional<>`, hence they need to be handled with some "care" inside of `Display`.

There's an abundance of attributes inside of all these structures. Writing all the Display implementation would take some time (and maintenance), plus I fear we might end up with an output that is just messy...

I changed my mind, I think we should go ahead with what the author proposes: use the json format inside of the `Display` implementations

flavio:
Closing in favor or https://github.com/krustlet/oci-distribution/pull/32

PR #29, opened 2022-04-06
Remove digest calculation of ImageData

tofay:
it previously calculated the digest of the image layers, by taking a digest of all the layers data. This calculation is not used in the OCI registry API, nor is the digest of an image.

(I intended to make this change in https://github.com/krustlet/oci-distribution/pull/22, as described in that PR description, but I didn't actually make the change...)

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -50,27 +50,6 @@ pub struct ImageData {
     pub manifest: Option<OciImageManifest>,
 }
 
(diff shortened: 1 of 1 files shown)


flavio (approved):

thomastaylor312 (approved):
LGTM! 

@flavio feel free to merge if you are good with everything

PR #33, opened 2022-05-04
Workaround AWS ECR issues

flavio:
Workaround push issues affecting AWS ECR
    
It seems AWS ECR is not respecting the OCI Distribution Spefication about:

* Push blobs as chunks: successful chunk uploads have return code 201 instead of 202
* Push Manifest: successful response is missing the `Location` header

This PR tries to work around this limitations, until the issues are sorted out by AWS.

These are the mitigations used.

* Wrong response with chunked uploads: when a spec violation is found, attempt the upload using the monolithic approach. ECR is respecting the spec in that regard.
* Missing `Location` header after manifest upload: given the whole image is already inside of the remote registry, we calculate the `Location` value by using the digest of the manifest. This is the value that is expect to be returned by the remote registry.

In both cases, warning messages are emitted to inform the user about the violations.

This PR introduces also some example programs that can be used to push and pull Wasm modules to/from OCI registry.

~~Note well: this PR is based on top of https://github.com/krustlet/oci-distribution/pull/31~~ Update: the PR got merged

### DIFF
--- a/examples/wasm/pull.rs
+++ b/examples/wasm/pull.rs
@@ -1,0 +1,24 @@
+use oci_distribution::{manifest, secrets::RegistryAuth, Client, Reference};
+use tracing::info;
+
+pub(crate) async fn pull_wasm(
+    client: &mut Client,
+    auth: &RegistryAuth,
+    reference: &Reference,
+    output: &str,
+) {
+    info!(?reference, ?output, "pulling wasm module");
+
+    let image_content = client
+        .pull(reference, auth, vec![manifest::WASM_LAYER_MEDIA_TYPE])
+        .await
+        .expect("Cannot pull Wasm module")
+        .layers
+        .into_iter()
+        .next()
+        .map(|layer| layer.data)
+        .expect("No data found");
+
+    std::fs::write(output, image_content).expect("Cannot write to file");
+    println!("Wasm module successfully written to {}", output);
+}
--- a/examples/wasm/pull.rs
+++ b/examples/wasm/pull.rs
@@ -19,6 +19,8 @@ pub(crate) async fn pull_wasm(
         .map(|layer| layer.data)
         .expect("No data found");
 
-    std::fs::write(output, image_content).expect("Cannot write to file");
+    async_std::fs::write(output, image_content)
+        .await
+        .expect("Cannot write to file");
     println!("Wasm module successfully written to {}", output);
 }
--- a/examples/wasm/push.rs
+++ b/examples/wasm/push.rs
@@ -14,7 +14,9 @@ pub(crate) async fn push_wasm(
 ) {
     info!(?reference, ?module, "pushing wasm module");
 
-    let data = std::fs::read(module).expect("Cannot read Wasm module from disk");
+    let data = async_std::fs::read(module)
+        .await
+        .expect("Cannot read Wasm module from disk");
 
     let layers = vec![ImageLayer::new(
         data,
--- a/examples/wasm/push.rs
+++ b/examples/wasm/push.rs
@@ -1,0 +1,36 @@
+use oci_distribution::{
+    client::{Config, ImageLayer},
+    manifest,
+    secrets::RegistryAuth,
+    Client, Reference,
+};
+use tracing::info;
+
+pub(crate) async fn push_wasm(
+    client: &mut Client,
+    auth: &RegistryAuth,
+    reference: &Reference,
+    module: &str,
+) {
+    info!(?reference, ?module, "pushing wasm module");
+
+    let data = std::fs::read(module).expect("Cannot read Wasm module from disk");
+
+    let layers = vec![ImageLayer::new(
+        data,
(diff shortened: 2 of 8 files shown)


flavio:
The tests are broken right now, because I didn't spend time fixing them. I would like to know from @bacongobbler and @thomastaylor312 if you are against introducing these workloads.

I've reached out to someone from AWS in the meantime to report the issue.

flavio:
> I've reached out to someone from AWS in the meantime to report the issue.

Update: they confirmed the issue. They are tracking it on their backlog.

In the meantime, I would like to go ahead and merge this PR. I think it's the good compromise between rejecting something that is non conformant and providing a good user experience

flavio:
rebased, since custom errors PR got merged

examples/wasm/push.rs
```diff
@@ -0,0 +1,36 @@
...
+
+pub(crate) async fn push_wasm(
+    client: &mut Client,
+    auth: &RegistryAuth,
+    reference: &Reference,
+    module: &str,
+) {
+    info!(?reference, ?module, "pushing wasm module");
+
+    let data = std::fs::read(module).expect("Cannot read Wasm module from disk");
```
thomastaylor312, line 17: Nit (because this is just an example): This is blocking IO in an async function. An async compatible read should be used instead

examples/wasm/pull.rs
```diff
@@ -0,0 +1,24 @@
...
+        .pull(reference, auth, vec![manifest::WASM_LAYER_MEDIA_TYPE])
+        .await
+        .expect("Cannot pull Wasm module")
+        .layers
+        .into_iter()
+        .next()
+        .map(|layer| layer.data)
+        .expect("No data found");
+
+    std::fs::write(output, image_content).expect("Cannot write to file");
```
thomastaylor312, line 22: Nit (because this is just an example): This is blocking IO in an async function. An async compatible read should be used instead

src/client.rs
```diff
@@ -252,7 +252,7 @@ impl Client {
     pub async fn push(
         &mut self,
         image_ref: &Reference,
-        layers: Vec<ImageLayer>,
+        layers: &[ImageLayer],
```
thomastaylor312, line 255: Super nit: Because this is a bug fix PR, I'd prefer to avoid the public API break here.

However, this is probably a good API change to make, so not going to block on it as we are coming up on a new release anyway

src/client.rs
```diff
@@ -461,6 +503,11 @@ impl Client {
                 reqwest::StatusCode::UNAUTHORIZED => {
                     Err(OciDistributionError::UnauthorizedError { url })
                 }
+                s if s.is_success() => Err(OciDistributionError::SpecViolationError(format!(
```
thomastaylor312, line 506: Note to our future selves: looks like our status error handling code needs to be put into a single function somewhere

thomastaylor312 (approved):
Yay for non spec compliant things! This all looks LGTM. There are a few nits, but no need to address them.

thomastaylor312:
Noting this as breaking due to the API changes so we don't forget to call them out in the release notes

src/client.rs
flavio: Good point, fixed

examples/wasm/push.rs
flavio: Good point, fixed

examples/wasm/pull.rs
flavio: Good point, fixed

src/client.rs
flavio: Good point. Done that too

ereslibre (commented):
Thanks @flavio, just a minor comment WRT to the error name.

src/client.rs
```diff
@@ -265,18 +265,41 @@ impl Client {
...
 
-        let config_url = self
-            .push_blob(image_ref, config.data, &manifest.config.digest)
-            .await?;
+        let config_url = match self
+            .push_blob_chunked(image_ref, &config.data, &manifest.config.digest)
+            .await
+        {
+            Ok(url) => url,
+            Err(OciDistributionError::SpecViolationError(violation)) => {
```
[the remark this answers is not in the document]
flavio: The errors is generated at a lower level, inside of shared code that is used by push/pull and other operations. Depending on the operation done, the remote registry has to respond with a limited set of HTTP status codes. Diverging from these status codes is what produced a `SpecViolationError`.

The error is then handled in different ways, depending on the context. For example, in this case, we are doing a chunked push and we know there are good chances that doing a single push would work.
In another piece of the code, we are not getting a HTTP header set, hence we react to the SpecViolationError in a different way.

Long story short, I don't think that creating a series of SpecViolationErrors is doable (the errors are raised at a lower portion of the call stack, where there's no knowledge of the context). `map`-ing the generic SpecViolationError can be done only at a higher point of the stack, in the places were we are already trying the workarounds.

Issue #30, opened 2022-04-12
Unexpected org.opencontainers.image.title in manifest


VandanaAkhramka:
The manifest created by `oci_distribution::client::Client::push` uses an incorrect annotation. It sets the `org.opencontainers.image.title` field as the value of the digest but it should actually save it as the name of the artifact that is being pushed, otherwise, when you pull it, the downloaded artifact's name is in the format `sha256:<string>` instead of the expected filename.

PR #34, opened 2022-05-05
Fix manifest annotations

flavio:
The main change introduced by this PR is to better support [OCI annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md):

* Define a series of constants, one per pre-defined OCI annotation. This one of our TODOs
* Extend `Config` and `Manifest` and other objects to allow them to have OCI annotations, as expressed by the OCI Image Spec
* Provide an helper constructor for `OciImageIndex`. That allows users to create a Manifest with annotations, that can then be passed to the `Client::push` method. This helper is obtained by converting a private helper function into a public one
* Do not set a default `org.opencontainers.image.title` annotation inside of the ImageLayers. This fixes issue #30
* Extend the `example wasm push` command to make use of these changes
    
This PR introduces 2 small changes that are unrelated with the main topic:
  * Fix links inside of inline docs: address the `cargo doc` warnins. A lot of our inline links were not rendered as hyperlinks, but as simple text
  * Small refactoring of some existing code: make use of an internal helper function to remove some duplicated code

**Note:** this PR builds on top of https://github.com/krustlet/oci-distribution/pull/33 I'll rebase it gets merged


src/lib.rs
```diff
@@ -4,6 +4,7 @@
 pub mod client;
 pub mod errors;
 pub mod manifest;
+pub mod oci_annotations;
```
thomastaylor312, line 7: Since this is called the `oci_distribution` crate, we should probably just call this module `annotations`. It also matches the rest of the module names that aren't prefaced by it

src/client.rs
```diff
@@ -1434,7 +1428,7 @@ fn digest_header_value(headers: HeaderMap, body: Option<&str>) -> Result<String>
 }
 
 /// Computes the SHA256 digest of a byte vector
-fn sha256_digest(bytes: &[u8]) -> String {
+pub(crate) fn sha256_digest(bytes: &[u8]) -> String {
```
thomastaylor312, line 1431: This probably belongs at the top level `lib.rs` as it is used in multiple places and doesn't have much to do with the client itself

thomastaylor312 (changes requested):
Super minor changes request here. Otherwise everything looks good

src/lib.rs
flavio: good point, done

src/client.rs
flavio: good point, done

flavio:
rebased

PR #39, opened 2022-05-26
WIP: docker config auth

imjasonh:
This adds a `RegistryAuth` option that instructs the client to find, parse, and use a Docker config file (typically at `~/.docker/config.json`) if possible.

If the config file contains a username and password (in either form) or an identity token, for the requested registry, it will be used.

If the config file specifies a cred helper for the given registry, the cred helper will be invoked and its output parsed to find a username and password (or ID token). (TODO: this is not implemented correctly now).

If any of the above fails, the error will be debug-logged and anonymous auth will be used instead.


src/secrets.rs
```diff
@@ -1,22 +1,143 @@
...
     /// Access the registry using HTTP Basic authentication
     Basic(String, String),
+
+    /// Access the registry using credentials configured in the Docker config file
+    Default,
 }
 
 pub(crate) trait Authenticable {
-    fn apply_authentication(self, auth: &RegistryAuth) -> Self;
+    fn apply_authentication(self, registry: &str, auth: &RegistryAuth) -> Self;
```
thomastaylor312, line 29: Instead of passing a registry, I'd recommend making `RegistryAuth::Default` be `RegistryAuth::Default(String)`. This means no API change and also makes sure the config information is part of the enum

src/secrets.rs
```diff
@@ -1,22 +1,143 @@
...
 /// A method for authenticating to a registry
 pub enum RegistryAuth {
     /// Access the registry anonymously
     Anonymous,
+
     /// Access the registry using HTTP Basic authentication
     Basic(String, String),
+
+    /// Access the registry using credentials configured in the Docker config file
+    Default,
```
thomastaylor312, line 25: Nit: Should probably rename `DefaultKeychain` to match what this method is called in other places like go

src/secrets.rs
```diff
@@ -1,22 +1,143 @@
...
+        let password = resp.secret;
+        return Ok((username, password));
     }
+
+    // TODO: credsStores
+
+    Err(anyhow::Error::msg("no matching credentials"))
 }
+
+fn get_docker_config() -> Result<DockerConfig, anyhow::Error> {
```
thomastaylor312, line 103: This should be `async` and use the `tokio::fs` functions instead as it is being used in an async context. Unfortunately async stuff isn't as easy to use as Go's is, so you have to use a different set of types that are meant for async

src/secrets.rs
```diff
@@ -1,22 +1,143 @@
...
+    // TODO: Also check for Podman auth configured in $XDG_RUNTIME_DIR/containers/auth.json
+    let cfg = env::var("DOCKER_CONFIG")
+        .unwrap_or(
```
thomastaylor312, line 106: Generally if you are calling a function, you should use `unwrap_or_else` so that the function is called only when it is needed

src/secrets.rs
```diff
@@ -1,22 +1,143 @@
...
         }
+        if ac.identity_token != "" {
+            return Ok(("<token>".to_string(), ac.identity_token.to_string()));
+        }
+        // TODO: bearer auth if RegistryToken is found.
+    }
+
+    // If a cred helper is configured, execute it, passing the registry to stdin and parsing its stdout as JSON.
+    let chr = cfg.cred_helpers.get(registry);
+    if !chr.is_none() {
```
thomastaylor312, line 83: Rust has the handy syntax of 

```rust
if let Some(ch) = chr {
    let out = Command::new
    .....
}
```

src/secrets.rs
```diff
@@ -1,22 +1,143 @@
...
+    if acr.is_none() && registry == "index.docker.io" {
+        acr = cfg.auths.get("https://index.docker.io/v1/")
+    }
+    if !acr.is_none() {
+        let ac = acr.unwrap();
+
+        if ac.auth != "" {
+            // base64-decode the auth to get username and password.
+            let dec = &base64::decode(&ac.auth)?;
+            let vec = str::from_utf8(dec)?.split(":").collect::<Vec<_>>();
```
thomastaylor312, line 66: There is a `split_once` function that you can use here so you don't have to check the length

src/secrets.rs
```diff
@@ -1,22 +1,143 @@
...
-    fn apply_authentication(self, auth: &RegistryAuth) -> Self {
+    fn apply_authentication(self, registry: &str, auth: &RegistryAuth) -> Self {
         match auth {
             RegistryAuth::Anonymous => self,
             RegistryAuth::Basic(username, password) => self.basic_auth(username, Some(password)),
+            RegistryAuth::Default => match docker_auth(registry) {
+                Ok(userpass) => self.basic_auth(userpass.0, Some(userpass.1)),
+                Err(e) => {
+                    // This effectively falls back to anonymous, instead of raising the error.
+                    // TODO: Raise this error so clients can *require* Docker config auth.
```
thomastaylor312, line 41: This is a good point. As we continue to add support for other auth methods, handling errors would be important. I think adjusting the trait to return `anyhow::Result<Self>` would be good

thomastaylor312 (commented):
Thank you for opening this! I offered some general rust pointers along with a few other comments. 

Let me know which TODOs can be captured as future enhancements and which ones should be resolved before we move this to ready for review

PR #40, opened 2022-05-30
Fix workaround used when pushing to ECR

flavio:
ECR currently doesn't properly follow the OCI distribution specification. Because of that, we had to implement a workaround when
performing manifest push.

Prior to this commit, the push manifest function returned the supposed digest of the OCI manifest when pushing to an ECR. However, the function was expected to return the URL to the immutable manifest location.

This commit fixes the behaviour of the workaround code, making it return a proper URL instead of just the manifest digest.


flavio:
Once this is merged I'll tag a new patch release

jvanz (commented):
LGTM

jvanz (approved):

PR #44, opened 2022-09-29
cleanups

flavio:
This PR performs a couple of cleanups.

## Get rid of stale dependencies

Remove the following dependencies:
* hyperx
* www-authenticate

The `hyperx` crate doesn't seem to be maintained anymore. Its dependencies have upper bounds that are causing upgrade conflicts both to oci-distribution and to all the consumers of it.
See https://github.com/dekellum/hyperx/issues/39

hyperx is a fork of a really old version of the popular hyper crate. We require it just to be able to consume the `www-authenticate` crate, which depends on types that are defined inside of hyperx. These types have been removed from the official hyper crate.

This commit replaces these two dependencies with a more modern one called `http-auth`, which builds on top of the popular `http` crate.

This commit also extends the test suite to perform checks against a docker registry that uses basic authentication.

No regression have been introduced by the refeactoring.

## Remove the `url` dependency

Remove the `url` dependency, use the re-export from `reqwest`.

## Update dependencies

Update dependencies to latest stable versions.

## Enable dependabot

Run dependabot weekly to update both cargo dependencies and the github actions being used.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -1,7 +1,10 @@
 [advisories]
 ignore = [
     # We are using some deprecated versions to support Windows
-    "RUSTSEC-2020-0016"
+    "RUSTSEC-2020-0016",
+
+    # a chrono issue, this is just a test dependency
+    "RUSTSEC-2020-0071",
 ]
 
 [licenses]
--- a/src/client.rs
+++ b/src/client.rs
@@ -19,7 +19,8 @@ use crate::errors::{OciDistributionError, Result};
 use crate::token_cache::{RegistryOperation, RegistryToken, RegistryTokenType, TokenCache};
 use futures_util::future;
 use futures_util::stream::StreamExt;
-use hyperx::header::{Header, Raw};
+use http::HeaderValue;
+use http_auth::{parser::ChallengeParser, ChallengeRef};
 use olpc_cjson::CanonicalFormatter;
 use reqwest::header::HeaderMap;
 use reqwest::{RequestBuilder, Url};
@@ -29,7 +30,6 @@ use std::collections::HashMap;
 use std::convert::TryFrom;
 use tokio::io::{AsyncWrite, AsyncWriteExt};
 use tracing::{debug, trace, warn};
-use www_authenticate::{Challenge, ChallengeFields, RawChallenge, WwwAuthenticate};
 
 const MIME_TYPES_DISTRIBUTION_MANIFEST: &[&str] = &[
     IMAGE_MANIFEST_MEDIA_TYPE,
@@ -409,15 +409,10 @@ impl Client {
             None => return Ok(()),
         };
 
-        let auth = WwwAuthenticate::parse_header(&Raw::from(dist_hdr.as_bytes()))
-            .map_err(|e| OciDistributionError::GenericError(Some(e.to_string())))?;
-        // If challenge_opt is not set it means that no challenge was present, even though the header
-        // was present.
-        let challenge_opt = match auth.get::<BearerChallenge>() {
-            Some(co) => co,
-            None => {
-                // Fall back to HTTP Basic Auth
-                debug!("Falling back to HTTP Basic Auth");
+        let challenge = match BearerChallenge::try_from(dist_hdr) {
+            Ok(c) => c,
+            Err(e) => {
+                debug!(error = ?e, "Falling back to HTTP Basic Auth");
                 if let RegistryAuth::Basic(username, password) = authentication {
                     self.tokens.insert(
                         image,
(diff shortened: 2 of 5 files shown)


src/client.rs
```diff
@@ -435,8 +430,7 @@ impl Client {
             RegistryOperation::Push => format!("repository:{}:pull,push", image.repository()),
         };
 
-        let challenge = &challenge_opt[0];
-        let realm = challenge.realm.as_ref().unwrap();
+        let realm = challenge.realm.as_ref();
```
viccuad, line 433: Do we have a mem leak from the `Box`ed realm now?

deny.toml
```diff
@@ -17,12 +17,6 @@ default = "deny"
 # [possible values: any SPDX 3.11 short identifier (+ optional exception)].
 allow = [
     "BSD-2-Clause",
-    "LicenseRef-krator",
-    "LicenseRef-krator-derive",
-    "LicenseRef-ring",
-    "LicenseRef-webpki",
-    "LicenseRef-webpki-roots",
-    "MPL-2.0",
```
viccuad, line 25: I'm curious on why these licenses are being removed from the allow list (granted that I have failed to find some, like `LicenseRef-krator`, in the [SPDX list](https://spdx.org/licenses/)).

viccuad (commented):
(Not a project reviewer) This looks good to me, thanks for the new tests and the cleanup!
Just a couple of questions.

deny.toml
flavio: I removed them because `cargo deny` was showing warning messages about these licenses not being used anymore. I guess they are there because the `deny.toml` was initially copied from the one of krustlet, which had way more dependencies.

However, I had to restore the old file because some licenses were actually being used when building on windows

src/client.rs
flavio: I don't think so... please correct me if I'm wrong.

`realm` holds a reference to a `&str` that is built starting from a value provided by a `ChallengeParser`. `ChallengeParser` has a lifetime equal to the value it's built against: the `HeaderValue`: https://github.com/krustlet/oci-distribution/blob/a7189f495aa9efebda878b95121f7e2981dee095/src/client.rs#L1375-L1379

The `HeaderValue` is created inside of the `auth` function, https://github.com/krustlet/oci-distribution/blob/a7189f495aa9efebda878b95121f7e2981dee095/src/client.rs#L407-L410

I think `realm` and `HeaderValue`all go away at the same time, once the scope of `auth` is left.

Am I missing something?

viccuad: Not at all, all fine! Got confused with removed code and thought we were doing an `into_raw()` instead of `into_boxed_str()` for its creation.

viccuad (approved):

src/client.rs
```diff
@@ -1366,41 +1360,76 @@ impl ClientProtocol {
     }
 }
 
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 struct BearerChallenge {
-    pub realm: Option<String>,
+    pub realm: Box<str>,
     pub service: Option<String>,
+    #[allow(dead_code)]
```
thomastaylor312, line 1367: Is there a reason we want to allow dead code? Is this for future use?

src/client.rs
```diff
@@ -1366,41 +1360,76 @@ impl ClientProtocol {
...
+        let parser = ChallengeParser::new(
+            value
+                .to_str()
+                .map_err(|e| format!("cannot convert header value to string: {:?}", e))?,
+        );
+        let challenge_opt: Vec<BearerChallenge> = parser
+            .filter_map(|parser_res| {
+                if let Ok(chalenge_ref) = parser_res {
+                    let bearer_challenge = BearerChallenge::try_from(&chalenge_ref);
+                    bearer_challenge.map_or_else(|_e| None, Some)
```
thomastaylor312, line 1384: This should just use `.ok()`

src/client.rs
```diff
@@ -1366,41 +1360,76 @@ impl ClientProtocol {
...
+                } else {
+                    None
+                }
+            })
+            .collect();
+
+        if challenge_opt.is_empty() {
+            Err("Cannot find Bearer challenge".to_string())
+        } else {
+            Ok(challenge_opt[0].clone())
```
thomastaylor312, line 1394: Nit: I would use `to_owned()` here instead as it more clearly expresses intent. At first glance I thought you were cloning when we didn't need it. You could also simplify the logic by replacing the `if...then` with:

```rust
challenge_opt.into_iter().nth(0).map(|s| s.to_owned()).ok_or_else(|| "Cannot find Bearer challenge".to_string())
```

thomastaylor312 (changes requested):
Some small discussion points and nits. Otherwise this is looking great! Thank you for figuring this all out

src/client.rs
flavio: I think we can remove it, this is not being used. It wasn't used before either, I just kept it around. I removed it now

src/client.rs
flavio: done

src/client.rs
flavio: done

flavio:
@thomastaylor312 thanks for the review, I've applied all your suggestions

thomastaylor312 (approved):

Issue #53, opened 2022-12-08
Add stream processing API for pull_blob


arronwy:
current `pull_blob` API need pass an `AsyncWrite` like File or Memory. Image blobs are compressed/encrypted by default, to reduce memory and extra disk IO consumption, it can returns an `AsyncRead` which can integrate seamless with following blob processing using Rust's asynchronous IO types like `async_compression` https://docs.rs/async-compression/latest/async_compression/

PR #54, opened 2022-12-08
Add async_read support for pull_blob API

arronwy:
This API allows blobs input stream integrate seamless with following blob processing using Rust's asynchronous IO types like async_compression.

Fixed: #53

Signed-off-by: Wang, Arron <arron.wang@intel.com>


flavio (approved):
LGTM

flavio:
@arronwy I've just published v0.9.4 that features your patch

arronwy:
> @arronwy I've just published v0.9.4 that features your patch

Thanks a lot for your quick review and new v0.9.4 release, much appreciated!

thomastaylor312:
@flavio Not a big deal for something this small, but generally a feature add should probably be a new minor release rather than patch (so 0.10.0)

src/client.rs
```diff
@@ -775,6 +777,28 @@ impl Client {
...
 
+    /// Stream a single layer from an OCI registry.
+    ///
+    /// This is a streaming version of [`Client::pull_blob`].
+    /// Returns [`AsyncRead`](tokio::io::AsyncRead).
+    pub async fn async_pull_blob(
+        &self,
+        image: &Reference,
+        digest: &str,
+    ) -> Result<impl AsyncRead + Unpin> {
```
thomastaylor312, line 788: I know this is already merged but @arronwy is there a reason you chose `AsyncRead` over returning `Stream`? Generally it is better to return a stream and let the user choose how they want to consume it

arronwy: Hi @thomastaylor312 ， yes, agree, `Stream` can easy convert to `AsyncRead` I also verified if we return 
`Result<impl Stream<Item = std::result::Result<bytes::Bytes, std::io::Error>>>`
will also works. 

The reason I chose `AsyncRead` is try to consistent with current `pull_blob` API with output using:
`AsyncWrite` and `async_compression` crates: https://docs.rs/async-compression/latest/async_compression/index.html also operate over `AsyncWrite` and `AsyncBufRead` streams.

It's ok for me if we decide change to return `Stream` to let user to choose :-).

thomastaylor312: Right I had forgotten we used `AsyncRead`. I'll capture an issue to refactor that at some point

PR #56, opened 2022-12-14
fix: rustls allow usage or system root certificates

flavio:
Expose a feature of the reqwest crate that allows to use the system certificates.


viccuad (approved):
LGTM, tested it with the provided example against a local registry.

Issue #50, opened 2022-11-10
Support OCI Image Configuration 


Xynnn007:
Now the `Config` is only a poor `Vec<u8>` named `data`.
It is needed to support OCI Image Configuration due to https://github.com/opencontainers/image-spec/blob/v1.0/config.md.
`v1.1` is not selected because it is still in `rc` stage and not stable.

thomastaylor312:
Hey there @Xynnn007! Could you clarify a bit what you'd like to add here? As you mentioned, we have to do this because of the spec, were you suggesting we add something more structured? If so, what would you like it to look like?

Xynnn007:
Hey @thomastaylor312 , nice to get your feedback. As you've mentioned, I suggest adding something more structured as the spec defines  https://github.com/opencontainers/image-spec/blob/v1.0/config.md. 


I'm not an expert in this, but when I'm trying to test using a local registry, a pushed image in the registry contains a __more structured__ `Config`. A random structure of `Config` might let a (commercial) registry to clear the image.

Xynnn007:
I pushed a PR including definitions of the structures, but still, we need related functions?

lswith:
I've added 2 PRs related to this: 

- https://github.com/krustlet/oci-distribution/pull/77
- https://github.com/krustlet/oci-distribution/pull/76

I'm hoping that we can capture an "oci config" in the `ConfigFile` type, and convert from the `client::Config` to the `config::ConfigFile` and back. This will make it easier to parse images that have been pulled from a registry as well as generate new configs when pushing.

PR #57, opened 2022-12-15
feat: Add OCI Image Configuration structures

Xynnn007:
Related to https://github.com/krustlet/oci-distribution/issues/50, added structure definitions.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,0 +1,418 @@
+//! OCI Image Configuration
+//!
+//! Definition following <https://github.com/opencontainers/image-spec/blob/v1.0/config.md>
+
+use std::collections::{HashMap, HashSet};
+
+use chrono::{DateTime, Utc};
+use serde::{ser::SerializeMap, Deserialize, Deserializer, Serialize, Serializer};
+
+/// The CPU architecture which the binaries in this image are
+/// built to run on.
+/// Validated values are listed in [Go Language document for GOARCH](https://golang.org/doc/install/source#environment)
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Default)]
+#[serde(rename_all = "lowercase")]
+pub enum Architecture {
+    /// Arm
+    Arm,
+    /// Arm 64bit
+    Arm64,
+    /// Amd64/x86-64
+    #[default]
+    Amd64,
+    /// Intel i386
+    #[serde(rename = "386")]
+    I386,
+    /// Wasm
+    Wasm,
+    /// Loong64
+    Loong64,
+    /// MIPS
+    Mips,
+    /// MIPSle
+    Mipsle,
+    /// MIPS64
+    Mips64,
+    /// MIPS64le
+    Mips64le,
+    /// Power PC64
+    PPC64,
+    /// Power PC64le
+    PPC64le,
+    /// RiscV 64
+    Riscv64,
+    /// IBM s390x
+    S390x,
+    /// With this field empty
+    #[serde(rename = "")]
+    None,
+}
+
+/// The name of the operating system which the image is
+/// built to run on.
+/// Validated values are listed in [Go Language document for GOARCH](https://golang.org/doc/install/source#environment)
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Default)]
+#[serde(rename_all = "lowercase")]
+pub enum Os {
+    /// IBM AIX
+    Aix,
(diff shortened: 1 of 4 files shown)


src/config.rs
```diff
@@ -0,0 +1,399 @@
...
+        .ok_or_else(|| serde::de::Error::invalid_length(0, &"<TIME>"))?;
+
+    let ndt = NaiveDateTime::parse_from_str(s, "%+")
+        .map_err(|err| serde::de::Error::custom::<String>(err.to_string()))?;
+
+    Ok(Some(DateTime::<Utc>::from_utc(ndt, Utc)))
+}
+
+/// Helper to serialize a time
+fn serialize_time<S>(time: &Option<DateTime<Utc>>, s: S) -> Result<S::Ok, S::Error>
```
thomastaylor312, line 24: `chrono` has a `serde` feature that we should probably use rather than our own serializer/deserializer: https://docs.rs/chrono/latest/chrono/serde/index.html

src/config.rs
```diff
@@ -0,0 +1,399 @@
...
+    /// in the format `SIGNAME`, for instance `SIGKILL` or `SIGRTMIN+3`.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub stop_signal: Option<String>,
+}
+
+/// The rootfs key references the layer content addresses used by the image.
+/// This makes the image config hash depend on the filesystem hash.
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Default)]
+pub struct Rootfs {
+    /// MUST be set to `layers`.
```
thomastaylor312, line 214: We should probably have a `const` defined for `layers`

src/config.rs
```diff
@@ -0,0 +1,399 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub user: Option<String>,
+
+    /// A set of ports to expose from a container running this
+    /// image. Its keys can be in the format of: `port/tcp`, `port/udp`,
+    /// `port` with the default protocol being `tcp` if not specified.
+    /// These values act as defaults and are merged with any
+    /// specified when creating a container.
+    #[serde(skip_serializing_if = "HashMap::is_empty")]
+    pub exposed_ports: HashMap<String, serde_json::Value>,
```
thomastaylor312, line 174: The type it deserializes to in Go, which unfortunately influences what the spec looks like, is `map[string]struct{}`. So essentially it is acting as a set. So we have two options here:

1. The easier option (which I think I recommend for this PR): Make this a `HashMap<String, ()>` rather than a JSON value
2. The more Rust-y option (but requires more effort): Make this a `HashSet<String>` and then write a custom serializer/deserializer function for it

thomastaylor312 (changes requested):
A few small changes requested here.

How do you intend for this to be consumed? Does the user use the config type if they want it and then have to serialize it to `Vec<u8>`? Or is there another way to make it easier?

src/config.rs
Xynnn007: Yes, thanks. Fixed it

src/config.rs
Xynnn007: Nice suggestion, fixed it

src/config.rs
Xynnn007: I tried but it seems that `()` would be serialized into a `null` rather than a `{}`. Thus can not meet our needs. Another way I thought was to define an empty structure with no member fields. How does it like?

Xynnn007:
> How do you intend for this to be consumed? Does the user use the config type if they want it and then have to serialize it to `Vec<u8>`? Or is there another way to make it easier? Just want to make sure I understand!

I've added a new API `oci_v1_from_config_file` to construct a `Config` using a `ConfigFile`. Typical scenarios:
* A user wants to push a cosign signature and create a related structured Config
* A user wants to pull an image using `pull()` to get `ImageData`, and then he can parse the `image_data.config.data` into a structure to do some other things.

To me the first scenario is needed, because [current impl](https://github.com/sigstore/sigstore-rs/blob/main/src/cosign/client.rs#L140) uses [an empty config](https://github.com/sigstore/sigstore-rs/blob/main/src/cosign/client.rs#L34), which may be detected as illegal image for some registries. Thus, I want to provide some detailed support for OCI Configuration and let users use it arbitrarily.

thomastaylor312 (approved):
Thank you for all the effort with this! Everything is looking good

PR #61, opened 2023-01-23
Add support for compiling to wasm32 architecture

lulf:
Most of the dependencies like reqwest already handle compilation for wasm32, so it is a matter of setting only the supported tokio features and guard some of the reqwest APIs that are not supported on wasm32.

The underlying motivation is to use sigstore on wasm32 architectures, but being able to work with OCI registries from WASM modules might come in handy as well.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,7 @@ serde_json = "1.0"
 serde = { version = "1.0", features = ["derive"] }
 sha2 = "0.10"
 thiserror = "1.0"
-tokio = { version = "1.21", features = ["macros", "fs"] }
+tokio = { version = "1.21", features = ["macros", "io-util"] }
 tokio-util = { version = "0.7.4", features = ["compat"] }
 tracing = { version = "0.1", features = ['log'] }
 unicase = "2.6"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -49,7 +49,7 @@ serde_json = "1.0"
 serde = { version = "1.0", features = ["derive"] }
 sha2 = "0.10"
 thiserror = "1.0"
-tokio = { version = "1.21", features = ["macros", "fs"] }
+tokio = { version = "1.21", features = ["macros", "io-util"] }
```
thomastaylor312, line 52: Just double checking, we don't need the fs feature? I thought we were using that somewhere in the code

thomastaylor312 (commented):
Just one minor discussion item. Otherwise, this will be great!

Cargo.toml
lulf: Yeah, just removing the feature doesn't seem to affect anything on my local machine (and I guess neither in CI). 

The io-util feature is needed for some async traits that fails compilation with the wasm32 target.

flavio (approved):
Nice contribution. Thanks!

I trust the unit tests about the change of features for the `tokio` crate :)

thomastaylor312 (approved):

PR #63, opened 2023-02-07
Enhance client to push blobs, mount blobs, and push raw manifests

aochagavia:


src/client.rs
```diff
@@ -966,28 +961,62 @@ impl Client {
...
+    }
+
+    /// Pushes the manifest, provided as raw bytes, for a specified image
+    ///
+    /// Returns pullable manifest url
+    pub async fn push_manifest_raw(
+        &self,
+        image: &Reference,
+        body: Vec<u8>,
+        content_type: HeaderValue,
```
thomastaylor312, line 1007: This might be a cleaner API if this were to be `H` where `H: TryInto<HeaderValue, Error = OciDistributionError>` so people can pass anything that can be turned into a header value. Any thoughts on how you think this will be used?

thomastaylor312 (commented):
This is looking really great! Just one open question around the design of the API.

src/client.rs
```diff
@@ -966,28 +961,62 @@ impl Client {
...
 
+    /// Mounts a blob to the provided reference, from the given source
+    pub async fn mount_blob(
+        &self,
+        image: &Reference,
+        source: &Reference,
+        digest: &str,
+    ) -> Result<()> {
+        let base_url = self.to_v2_blob_upload_url(image);
+        let url = format!("{}?mount={}&from={}", base_url, digest, source.repository());
```
flavio, line 972: I might be a bit paranoid, but I would prefer to have the url built using the `Url` crate.

Using the [`query_pairs_mut`](https://docs.rs/url/2.3.1/url/struct.Url.html#method.query_pairs_mut) method would allow us to do proper encoding of the query parameters

flavio (commented):
Overall LGTM, I left a minor comment

thomastaylor312:
Just so others know what happened, we removed some large binaries from our git history and overwrote history, so this got automatically closed.

@aochagavia feel free to reopen a new PR against main when you're ready!

src/client.rs
aochagavia: I am using this function in an application that creates an OCI image in memory and pushes the result to the registry (see [this blog post](https://ochagavia.nl/blog/crafting-container-images-without-dockerfiles/) for details). The specific code is:

```rust
registry_client
    .push_manifest_raw(
        new_image_reference,
        image_manifest_json,
        OCI_IMAGE_MEDIA_TYPE.parse().unwrap(),
    )
    .await?;
```

I experimented changing the line you mentioned to `content_type: impl TryInto<HeaderValue, Error = OciDistributionError>` and ran into problems, but `impl TryInto<HeaderValue, Error = InvalidHeaderValue>` did the trick ;)

Issue #27, opened 2022-02-21
Support manifest list


thomasgouveia:
Hello,

First, thanks for your work on this crate, it is very helpful! 

I'm interested in the manifest list feature. Currently, I work on a crate to provide the ability to pull OCI images for a school project. So I need to be able to pull for example `alpine` or `ubuntu` images from `docker.io/library`.

For example, when I use the `pull_manifest_and_config` method of your `Client`, I have this error : 

`unsupported media type: application/vnd.docker.distribution.manifest.list.v2+json`

I checked your code and I see that the crates do not support the manifest list feature. Is this something under consideration? I see there is a TODO in your code, but no more information about future implementation. If not, did you know if there is a workaround to be able to pull those images? 

Thank you! 

Thomas

thomasgouveia:
Yes, I've looked #26, but I'm not sure it resolves my use case.  #26 is focused on the ability to push manifest lists, and not on pulling. What I'm interested in is the ability to pull images such as `docker.io/library/alpine` or `docker.io/library/ubuntu`.

To give you an idea, currently, with this piece of code : 

```rust
pub async fn pull_manifest(&mut self) -> Result<(OciManifest, String, String)> {
  self.client
      .pull_manifest_and_config(&self.reference, &self.auth)
      .await
      .map_err(|e| Error::PullManifest(e.to_string()))?
}
```

When my reference is parsed from `docker.io/library/ubuntu`, I have the following error : 

`unsupported media type: application/vnd.docker.distribution.manifest.list.v2+json`

So that's ok, as I said in my first comment.

But it is strange, because if I use `docker.io/amd64/ubuntu`, it works great.

I'm not sure to understand the difference between those two OCI references, and how the process has to be different for each.

PR #67, opened 2023-03-13
client: Fix failed to pull ubuntu image from docker.io

arronwy:
Manifests list feature in `oci-distribution` is already implemented, default image in docker.io like busybox use `IMAGE_MANIFEST_LIST_MEDIA_TYPE`, but ubuntu image use `OCI_IMAGE_INDEX_MEDIA_TYPE`.

Fixes: #27

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -696,6 +696,7 @@ impl Client {
             if media_type != IMAGE_MANIFEST_MEDIA_TYPE
                 && media_type != OCI_IMAGE_MEDIA_TYPE
                 && media_type != IMAGE_MANIFEST_LIST_MEDIA_TYPE
+                && media_type != OCI_IMAGE_INDEX_MEDIA_TYPE
             {
                 return Err(OciDistributionError::UnsupportedMediaTypeError(media_type));
             }


src/client.rs
```diff
@@ -2375,6 +2377,20 @@ mod test {
         assert_eq!(manifest.config.media_type, manifest::WASM_CONFIG_MEDIA_TYPE);
     }
 
+    #[tokio::test]
+    async fn test_pull_docker_ubuntu() {
+        let reference = Reference::try_from(DOCKER_IO_UBUNTU).expect("failed to parse reference");
```
flavio, line 2382: I'm afraid this test could report false negatives because of connectivity issues or rate limiting done by Docker Hub.

We are already starting a local instance of `distribution` during the execution of the unit tests. I think we should change this test to leverage this local instance, push something that uses a `OCI_IMAGE_INDEX_MEDIA_TYPE` inside of it and then try to pull it.

This would make the tests more robust

flavio (commented):
Thanks for the fix. I've left a comment about the tests

src/client.rs
arronwy: Yes, agree, we also have many rate limits issue during use docker hub.

Will fix, thanks a lot!

arronwy: I found local registry: https://github.com/distribution/distribution don't support `application/vnd.oci.image.index.v1+json` even it is defined in : https://github.com/opencontainers/image-spec/blob/main/media-types.md#compatibility-matrix
it will have following error when we try to push the manifest:
```
ERRO[1363] response completed with error   err.code="manifest invalid" 
err.detail="mediaType in manifest list should be 'application/vnd.docker.distribution.manifest.list.v2+json' not 'application/vnd.oci.image.index.v1+json'" err.message="manifest invalid" go.version=go1.20 http.request.contenttype=application/vnd.docker.distribution.manifest.list.v2+json 
http.request.host="localhost:5005" http.request.id=331dff94-3942-47be-803a-c0c18c1b1e8d 
http.request.method=PUT http.request.remoteaddr="127.0.0.1:39880" http.request.uri=/v2/ubuntu/manifests/latest http.request.useragent= http.response.contenttype=application/json http.response.duration="126.053µs" http.response.status=400 http.response.written=82 instance.id=2b9ba786-7076-4a8a-a5f9-b87a7f720225 service=registry vars.name=ubuntu vars.reference=latest version=v2.7.0-2130-ge5d58108
```

arronwy: submitted the registry issue in: https://github.com/distribution/distribution/issues/3863

flavio:
@arronwy Given I don't know how much time it will take for distribution to fix this issue, and given how easy the fix is on our side I propose the following approach:

* We create an issue to keep track of the missing unit test
* We merge this fix without any unit test

What do you think about that?

arronwy:
> @arronwy Given I don't know how much time it will take for distribution to fix this issue, and given how easy the fix is on our side I propose the following approach:
> 
> * We create an issue to keep track of the missing unit test
> * We merge this fix without any unit test
> 
> What do you think about that?

Agree, thanks a lot!

arronwy:
I have updated this PR, and create an issue to track uni tests: https://github.com/krustlet/oci-distribution/issues/68

PR #71, opened 2023-03-24
Change the visibility of `OciEnvelop.errors` to `pub`

linyinfeng:
The visibility of the registry error code (`OciErrorCode`) is `pub`, but there is no way to use it because the visibility of `OciEnvelop.errors` is `pub(crate)`.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -125,7 +125,8 @@ impl std::fmt::Display for OciError {
 /// A struct that holds a series of OCI errors
 #[derive(serde::Deserialize, Debug)]
 pub struct OciEnvelope {
-    pub(crate) errors: Vec<OciError>,
+    /// List of OCI registry errors
+    pub errors: Vec<OciError>,
 }
 
 impl std::fmt::Display for OciEnvelope {


radu-matei (approved):
Yeah, there are a few other such scenarios that we might want to make public.

This LGTM, thanks for your contribution!

flavio (approved):

src/errors.rs
```diff
@@ -125,7 +125,7 @@ impl std::fmt::Display for OciError {
 /// A struct that holds a series of OCI errors
 #[derive(serde::Deserialize, Debug)]
 pub struct OciEnvelope {
-    pub(crate) errors: Vec<OciError>,
+    pub errors: Vec<OciError>,
```
flavio, line 128: Can you add some inline documentation please? This is required by our test suite

flavio (changes requested):

src/errors.rs
linyinfeng: Sorry, I forgot that. Is this OK?

```rust
/// A struct that holds a series of OCI errors
#[derive(serde::Deserialize, Debug)]
pub struct OciEnvelope {
    /// List of OCI registry errors
    pub errors: Vec<OciError>,
}
```

flavio: Sure! :heart:

flavio (approved):

PR #72, opened 2023-03-24
make pushes concurrent and limit pull concurrency

dicej:
Previously, pushes were serialized and pulls were unboundedly concurrent.  This makes pushes concurrent (capped at 16 concurrent uploads) and also caps download concurrency at 16.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -18,9 +18,8 @@ use crate::Reference;
 
 use crate::errors::{OciDistributionError, Result};
 use crate::token_cache::{RegistryOperation, RegistryToken, RegistryTokenType, TokenCache};
-use futures::stream::TryStreamExt;
 use futures_util::future;
-use futures_util::stream::StreamExt;
+use futures_util::stream::{self, StreamExt, TryStreamExt};
 use http::HeaderValue;
 use http_auth::{parser::ChallengeParser, ChallengeRef};
 use olpc_cjson::CanonicalFormatter;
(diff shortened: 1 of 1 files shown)


bacongobbler:
Thanks @dicej!

Do you think it'd be useful if the maximum number of concurrent pulls/pushes was dependeny on the system hardware? For example, a Raspberry Pi may want to serialize less concurrent pulls/pushes, while a more beefy dev workstation may want to increase that count.

Most programs make this configurable by reading [the number of available CPUs](https://docs.rs/num_cpus/latest/num_cpus/) and provide an option to modify the default in case there are specific use cases (throttling/benchmarking).

What do you think about that approach?

bacongobbler:
Hmm. Perhaps it could be an optional field in the [`ClientConfig`](https://github.com/krustlet/oci-distribution/blob/f44124c3c0875821d0b84a6632bf70b6d6eaf9ef/src/client.rs#L1274) for now, defaulting to `MAX_PARALLEL_PUSH_AND_PULL` if unset. Would you agree with that approach for now?

bacongobbler:
I wonder if we need to make two separate variables: one for uploads (`MAX_PARALLEL_PUSH`) and one for downloads (`MAX_PARALLEL_PULL`).

dicej:
Yeah, that sounds great.  Update coming shortly.

bacongobbler (approved):

flavio:
Overall ok, but I would have preferred to have these configuration knobs added to the `ClientConfig` struct like @bacongobbler suggested. In this way we would not have needed the new methods of `Client` to be added.

dicej:
Ah, good point.  I misread @bacongobbler's comment.  I'll address that when I'm back at my keyboard.

dicej:
I've pushed an update.  Note that this required a version bump due to a breaking API change (new fields on the `ClientConfig` struct mean client code which construct instances of that struct will now need to specify those new fields).

flavio (approved):
LGTM, thanks for the change to the previous implementation proposal

PR #76, opened 2023-04-22
Conversion between a Config and a ConfigFile

lswith:
This PR implements the TryFrom for a client::Config so that it can be parsed back into a ConfigFile.

This feature is useful when pulling images from a registry, so that the metadata recieved can be viewed.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -90,6 +90,9 @@ pub enum OciDistributionError {
     /// Versioned object: JSON deserialization error
     #[error("Failed to parse manifest: {0}")]
     VersionedParsingError(String),
+    #[error(transparent)]
+    /// Transparent wrapper around `std::string::FromUtf8Error`
+    FromUtf8(#[from] std::string::FromUtf8Error),
 }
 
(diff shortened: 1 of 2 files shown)


src/errors.rs
```diff
@@ -90,6 +90,9 @@ pub enum OciDistributionError {
     /// Versioned object: JSON deserialization error
     #[error("Failed to parse manifest: {0}")]
     VersionedParsingError(String),
+    #[error(transparent)]
+    /// Transparent wrapper around `std::string::FromUtf8Error`
+    FromUtf8(#[from] std::string::FromUtf8Error),
```
flavio, line 95: I think it would be better to create a dedicated error for the Config parsing errors and ensure the `try_from` defined inside of this PR always returns this type of error.

flavio (commented):
I'm in favor of this change, I just have some questions about the error being raised by the new `try_from`.

@bacongobbler, @thomastaylor312: what do you think about that?

flavio:
I've rebased against the main branch and addressed some conflicts

flavio:
merged, sorry for the delay

PR #77, opened 2023-04-22
Update the configfile to match the oci v1 spec

lswith:
This PR updates the serialize/deserialize of the configfile to match the spec outlined [here](https://github.com/opencontainers/image-spec/blob/v1.0/config.md).

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -192,28 +192,30 @@ pub struct Config {
     #[serde(
         skip_serializing_if = "is_option_hashset_empty",
         deserialize_with = "optional_hashset_from_str",
-        serialize_with = "serialize_optional_hashset"
+        serialize_with = "serialize_optional_hashset",
+        default
     )]
     pub exposed_ports: Option<HashSet<String>>,
 
     /// Entries are in the format of `VARNAME=VARVALUE`.
-    #[serde(skip_serializing_if = "Vec::is_empty")]
-    pub env: Vec<String>,
+    #[serde(skip_serializing_if = "is_option_vec_empty")]
+    pub env: Option<Vec<String>>,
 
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -402,17 +437,140 @@ mod tests {
...
-    fn deserialize() {
+    fn deserialize_example() {
         let example = example_config();
         let parsed: ConfigFile = serde_json::from_str(EXAMPLE_CONFIG).expect("parsed failed");
         assert_eq!(example, parsed);
     }
 
     #[test]
-    fn serialize() {
+    fn deserialize_minimal() {
```
flavio, line 543: I think it would be nice to remove all this repetition by using the [`rstest`](https://crates.io/crates/rstest) library, like we are doing in other parts of the code.

I would propose to consolidate all these tests into a single one, something like:
```rust
#[rstest]
#[case(0, 0)]
#[case(1, 1)]
#[case(2, 1)]
#[case(3, 2)]
#[case(4, 3)]
fn fibonacci_test(#[case] name, #[case] input: &str, #[case] expected: ConfigFile) {
        let parsed: ConfigFile = serde_json::from_str(input).expect(format!("{name}: parsed failed").as_str());
        assert_eq!(expected, parsed, "something went wrong with test case {name}");
}
```

flavio (commented):
I'm in favor of changing the structs in this way, I'm referring to the usage of `Optional`.

I left some comments about minor changes

flavio:
@thomastaylor312 , @bacongobbler: what do you think about this change?

flavio:
I've rebased against `main` and fixed the conflicts

flavio:
merging, sorry for the delay

PR #79, opened 2023-04-22
Pass the image layers as a Vector

lswith:
This PR is the result of a long winding issue when calling the `client.push` function from
within `tokio::spawn`.

The specific error was:

```
error: implementation of `Iterator` is not general enough
  --> src/tester.rs:54:5
   |
54 |     res
   |     ^^^ implementation of `Iterator` is not general enough
   |
   = note: `Iterator` would have to be implemented for the type `std::slice::Iter<'0, ImageLayer>`, for any lifetime `'0`...
   = note: ...but `Iterator` is actually implemented for the type `std::slice::Iter<'1, ImageLayer>`, for some specific lifetime `'1`
```
and 
```
error: implementation of `FnOnce` is not general enough
  --> src/tester.rs:54:5
   |
54 |     res
   |     ^^^ implementation of `FnOnce` is not general enough
   |
   = note: closure with signature `fn(&'0 ImageLayer) -> [async block@Client::push::{closure#0}::{closure#0}::{closure#0}]` must implement `FnOnce<(&ImageLayer,)>`, for any lifetime `'0`...
   = note: ...but it actually implements `FnOnce<(&ImageLayer,)>
```

or 
```
error: higher-ranked lifetime error
   --> src/client.rs:92:5
    |
92  | /     async move {
93  | |         if *auth != RegistryAuth::Anonymous {
94  | |             client
95  | |                 .auth(&image, auth, oci_distribution::RegistryOperation::Push)
...   |
99  | |         client.push(&image, &layers, config, auth, manifest).await
100 | |     }
    | |_____^

warning: `registry-load-tester` (lib) generated 2 warnings
error: could not compile `registry-load-tester` due to previous error; 2 warnings emitted
```

Only once I found this specific rust issue: https://github.com/rust-lang/rust/issues/89976 did it make sense that
this was the issue.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -281,7 +281,7 @@ impl Client {
         self.validate_layers(&manifest, accepted_media_types)
             .await?;
 
-        let layers = stream::iter(&manifest.layers)
+        let layers = stream::iter(manifest.layers.clone())
             .map(|layer| {
                 // This avoids moving `self` which is &mut Self
                 // into the async block. We only want to capture
@@ -322,7 +322,7 @@ impl Client {
     pub async fn push(
         &mut self,
         image_ref: &Reference,
-        layers: &[ImageLayer],
+        layers: Vec<ImageLayer>,
         config: Config,
         auth: &RegistryAuth,
         manifest: Option<OciImageManifest>,
@@ -335,7 +335,7 @@ impl Client {
 
         let manifest: OciImageManifest = match manifest {
             Some(m) => m,
-            None => OciImageManifest::build(layers, &config, None),
+            None => OciImageManifest::build(layers.as_ref(), &config, None),
         };
 
         // Upload layers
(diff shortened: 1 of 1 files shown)


src/client.rs
```diff
@@ -281,7 +281,7 @@ impl Client {
         self.validate_layers(&manifest, accepted_media_types)
             .await?;
 
-        let layers = stream::iter(&manifest.layers)
+        let layers = stream::iter(manifest.layers.clone())
```
lswith, line 284: This is just a clone of the descriptors, not the actual layers.

flavio:
Can you show a snippet of the code under `tester.rs` that triggered the error?

lswith:
I can't seem to find the code or replicate so I'm happy to remove this PR for now and if it comes back, I'll make a better example.

PR #81, opened 2023-05-01
Update default `docker.io` registry

radu-matei:
This commit updates the default endpoint for DockerHub to be
`index.docker.io`. This is the default endpoint that the Docker CLI is
using.

Running `docker info`, this is the endpoint returned for the registry:

```
$ docker info | grep Registry
 Registry: https://index.docker.io/v1
```

This means we can now push references to
`index.docker.io/user/repo:tag`.

Signed-off-by: Radu Matei <radu.matei@fermyon.com>

### DIFF
--- a/src/reference.rs
+++ b/src/reference.rs
@@ -108,7 +108,7 @@ impl Reference {
     pub fn resolve_registry(&self) -> &str {
         let registry = self.registry();
         match registry {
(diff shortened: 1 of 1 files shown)


radu-matei:
(CI failure is not related to the changes in this PR.)

flavio (approved):

Issue #58, opened 2022-12-15
Swap out `AsyncRead`/`AsyncWrite` for `Stream`


thomastaylor312:
Both [`pull_blob`](https://github.com/krustlet/oci-distribution/blob/abc971315c421bde87ab6c40c88caffccb2d4c8f/src/client.rs#L758) and [`async_pull_blob`](https://github.com/krustlet/oci-distribution/blob/abc971315c421bde87ab6c40c88caffccb2d4c8f/src/client.rs#L784) should probably be taking and returning `impl Stream` rather than read or write. That should allow more flexibility for people to consume the stream how they wish.

Also, now that I am reading this again, I think that we should rename `async_pull_blob` to `pull_blob_stream`

PR #83, opened 2023-05-21
client: Return Stream for async_pull_blob API

arronwy:
Return Stream and allow the user to decide how to use it.
Move dependency of `tokio-util` to dev for tests.

Fixes: #58


flavio (commented):
LGTM. Can you look into the failing check please?

flavio:
> LGTM. Can you look into the failing check please?

Scratch that, the CI was failing because of unrelated issues. I'm fixing them here https://github.com/krustlet/oci-distribution/pull/84

PR #89, opened 2023-07-13
add optional artifactType to image manifest for oci v1.1

devigned:
Add an optional `artifactType` field to the image manifest to reflect the new changes for OCI v1.1.

See: https://github.com/opencontainers/image-spec/blob/main/manifest.md#image-manifest


flavio (approved):
Thanks for the contribution

devigned:
Thank you for the review and merge, @flavio!

PR #93, opened 2023-09-16
Improved Push Logic

rbbl-dev:
* Added Media Type Extraction to OciManifest.content_type() to facilitate pushing of Docker Images
* Updated push_manifest to be public cause i believe that its generally useful for library consumers. [Usage Example](https://gitlab.com/rbbl/devops-tools/semtag/-/blob/main/src/main.rs?ref_type=heads)

### DIFF
--- a/src/manifest.rs
+++ b/src/manifest.rs
@@ -51,10 +51,16 @@ pub enum OciManifest {
 
 impl OciManifest {
     /// Returns the appropriate content-type for each variant.
-    pub fn content_type(&self) -> &str {
+    pub fn content_type(&self) -> String {
         match self {
-            OciManifest::Image(_) => OCI_IMAGE_MEDIA_TYPE,
-            OciManifest::ImageIndex(_) => IMAGE_MANIFEST_LIST_MEDIA_TYPE,
+            OciManifest::Image(image) => image
+                .clone()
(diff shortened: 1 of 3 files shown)


src/manifest.rs
```diff
@@ -51,10 +51,16 @@ pub enum OciManifest {
 
 impl OciManifest {
     /// Returns the appropriate content-type for each variant.
-    pub fn content_type(&self) -> &str {
+    pub fn content_type(&self) -> String {
```
jsoverson, line 54: This change is better off as below to avoid the clone and the signature change

```rust
    pub fn content_type(&self) -> &str {
        match self {
            OciManifest::Image(image) => {
                image.media_type.as_deref().unwrap_or(OCI_IMAGE_MEDIA_TYPE)
            }
            OciManifest::ImageIndex(image) => image
                .media_type
                .as_deref()
                .unwrap_or(IMAGE_MANIFEST_LIST_MEDIA_TYPE),
        }
    }
```

jsoverson:
I don't know about the implications of the `media_type` change, but I was just about to submit a PR to make `push_manifest()` public so +1 to getting that in.

flavio (approved):
thanks for the contribution

Issue #97, opened 2023-10-10
Add support for the signatures registry API extension


mattarnoatibm:
OpenShift and IBM Cloud container registries support a signatures API extension where Red Hat Simple Signing signatures signatures for an image digest can be stored in and fetched from the container registry.

We could add support for this extension to the OCI distribution client.

Under this issue we might also want to design how the oci-distribution Rust library should support extensions to the OCI distribution spec geerally.

Refs:
 - https://access.redhat.com/documentation/en-us/openshift_container_platform/3.10/html/cluster_administration/admin-guide-image-signatures#accessing-image-signatures-using-registry-api
 - https://github.com/containers/image/blob/main/docs/signature-protocols.md#openshift-dockerdistribution-api-extension

thomastaylor312:
So I am not personally familiar with the RSS extension, but I am definitely not opposed to adding in registry extensions. The main thing I'd want to figure out before reviewing #98 is how we decide on which extensions to include. Given that this specific extension seems to be relatively well-known, it feels like it would fit, but I also don't want to get to a point where we include every extension anyone would ever use

@flavio I'm curious if you have any thoughts here on how we should weigh extensions for inclusion in the main crate or if there is another way we could possibly handle them

thomastaylor312:
I don't think I see extensions here https://github.com/google/go-containerregistry/tree/v0.16.1/pkg/v1 or here: https://pkg.go.dev/github.com/docker/docker@v24.0.6+incompatible/distribution, so I am not quite certain how this is done in other clients

flavio:
I didn't know this mechanism of registry API extensions. From what I understand this is something specific to the registry used by OpenShift. @mattarnoatibm am I right?

I have mixed feeling about getting this code into the main crate. Right now the code from the PR is pretty small and enabled only via a new dedicated feature flag.

The PR adds support for pulling and validating the signatures, but it doesn't seem to allow the creation/push of these objects. Am I right? I wonder how big this specific code could become in the future.

As for having this feature into a separate crate. I looked at the code added by #98. I think this could be moved to a different crate. We would have to extend the `Client::auth` return signature to return the authentication token.
All the other operations done by the PR can be performed with a vanilla `reqwest.Client` and the auth token.

What do you think?

thomastaylor312:
That seems to be the simplest way forward to me and avoids taking on all the code into this crate. As things evolve, we could also make this easier by creating some sort of trait that other crates could implement

[...]

PR #98, opened 2023-10-13
feat: add support for signatures api extension

mattarnoatibm:
Add support for the signatures registry API extension so for an image digest, a simple signing signature can be fetched from a registry that supports the extension.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ maintenance = {status = "actively-developed"}
 
 [features]
 default = ["native-tls", "test-registry"]
+extension-rss = []
 native-tls = ["reqwest/native-tls"]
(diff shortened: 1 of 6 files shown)


mattarnoatibm:
If the github workflow has access to a credential for fetching the manifest or signature from the repository in the uk.icr.io registry, which requires authentication, the tests that use the uk.icr.io registry will run.

If the github workflow does not have access to this credential the tests will not run.

mattarnoatibm:
Linked issue - https://github.com/krustlet/oci-distribution/issues/97

flavio:
This can be closed since we decided to use this approach instead: https://github.com/krustlet/oci-distribution/pull/105

PR #99, opened 2023-10-20
feat(config.rs): add OS type for WASI Preview 1

vdice:
Re: https://go.dev/doc/install/source#environment and https://go.dev/blog/wasi


thomastaylor312 (approved):
We should have added this a while ago. Thank you @vdice!

PR #105, opened 2023-11-23
feat: return auth token from Client::auth

mattarnoatibm:
Return the auth token from `Client::auth` if the container registry supports bearer token authentication.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -498,7 +498,7 @@ impl Client {
         image: &Reference,
         authentication: &RegistryAuth,
         operation: RegistryOperation,
-    ) -> Result<()> {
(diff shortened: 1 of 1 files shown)


flavio (approved):
This looks good to me.

@mattarnoatibm, I guess you've already verified that this small change is all you need to be unblocked. Am I right?

mattarnoatibm:
Hi Flavio. Thanks for looking at this PR.

Yes, I tested this change is sufficient using a feature branch of the guest-components/image-rs crate and the `signature_verification_xrss` test from that crate, which uses a custom container registry API extension (the get signatures one).

PR #107, opened 2023-12-06
Fix crate version: 0.11 -> 0.10

flavio:
The crate version was already bumped. We never released 0.10, hence there's no need to jump straight from 0.9 to 0.11.


flavio:
@thomastaylor312 I just noticed, before tagging 0.11, that we never had the 0.10 release.

thomastaylor312 (approved):
Whoops. Good catch.

PR #108, opened 2024-01-22
Pull alternative urls

seanyoung:
This helps with pulling this image
mcr.microsoft.com/windows/nanoserver/insider:10.0.20348.1


src/client.rs
```diff
@@ -855,25 +854,45 @@ impl Client {
...
+        let mut alternatives = if let Some(alternatives) = &layer.urls {
+            alternatives.iter()
+        } else {
+            [].iter()
+        };
+
+        let mut response;
+
+        while {
+            response = RequestBuilderWrapper::from_client(self, |client| client.get(url))
```
thomastaylor312, line 879: I did some digging here and I don't think we can fetch the external URLs in this way. If you look at the code in the [golang implementation](https://github.com/containers/image/blob/7d484406f74930df33869337d63bbe0b04592ce2/docker/docker_client.go#L980) it looks like we shouldn't be doing authorization with the same credentials since it can cause credential exposure (see [CVE-2020-15157](https://nvd.nist.gov/vuln/detail/CVE-2020-15157)).

src/client.rs
```diff
@@ -855,25 +854,45 @@ impl Client {
...
-            .apply_auth(image, RegistryOperation::Pull)?
-            .into_request_builder()
-            .send()
-            .await?
-            .error_for_status()?
-            .bytes_stream();
+        let mut url =
+            &self.to_v2_blob_url(image.resolve_registry(), image.repository(), &layer.digest);
+
+        let mut alternatives = if let Some(alternatives) = &layer.urls {
```
thomastaylor312, line 870: I think to be spec compliant, we should be skipping any urls that don't have a `http` or `https` scheme. See https://github.com/opencontainers/image-spec/blob/v1.0.1/descriptor.md#properties

thomastaylor312 (changes requested):
Thank you so much for adding this! Just have some follow ups after reading through the spec and other reference implementations

src/client.rs
```diff
@@ -855,25 +854,45 @@ impl Client {
...
+            alternatives.iter()
+        } else {
+            [].iter()
+        };
+
+        let mut response;
+
+        while {
```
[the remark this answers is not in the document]
seanyoung: Re-written to be a regular for loop - since the alternative urls should use authentication credentials

src/client.rs
seanyoung: Nice catch, fixed

src/client.rs
seanyoung: Fixed

thomastaylor312 (approved):
Thank you the quick response! This is good to go. Since this is a breaking API change, we'll release this in the 0.11 release

seanyoung:
> Thank you the quick response! This is good to go. Since this is a breaking API change, we'll release this in the 0.11 release

Thanks! I've fixed the needless borrow warnings, needs approval again. I suspect the test failure was just an intermittent failure

src/client.rs
```diff
@@ -855,25 +854,49 @@ impl Client {
...
+                    // can be abused to leak credentials or tokens.  Please
+                    // refer to CVE-2020-15157 for more information.
+                    response =
+                        RequestBuilderWrapper::from_client(self, |client| client.get(url.clone()))
+                            .apply_accept(MIME_TYPES_DISTRIBUTION_MANIFEST)?
+                            .into_request_builder()
+                            .send()
+                            .await?
+                }
+            }
```
flavio, line 896: The OciDescriptor contains a list of urls from which the the object might be downloaded. Here we are iterating over all the urls and making multiple GET requests. We keep overwriting the response inside of the loop, meaning that on line `899` we look only into the response we sent to the last url we found.

Should we instead do something like:

- for `url` in `urls`
  - make connection against `url`
  - if response is success, break from the loop
- outside of the loop:
  - if everything failed -> raise an error ; else process the successful byte stream

src/client.rs
```diff
@@ -889,15 +912,42 @@ impl Client {
...
```
flavio, line 946: Same concern over there

src/client.rs
```diff
@@ -1478,6 +1528,18 @@ pub fn linux_amd64_resolver(manifests: &[ImageIndexEntry]) -> Option<String> {
         .map(|entry| entry.digest.clone())
 }
 
+/// A platform resolver that chooses the first windows/amd64 variant, if present
+pub fn windows_amd64_resolver(manifests: &[ImageIndexEntry]) -> Option<String> {
```
flavio, line 1532: Why do we need this public function? It doesn't seem to be used anywhere inside of the code

flavio (changes requested):

src/client.rs
seanyoung: > The OciDescriptor contains a list of urls from which the the object might be downloaded. Here we are iterating over all the urls and making multiple GET requests.

We stop iterating once we find the last response succeeded. That last response might be the main url constructed from the digest, hence the check for "did the last request succeed" is at the top of the loop.

We are only making multiple get requests if the previous one failed. 
 
> Should we instead do something like:
> 
>  * for `url` in `urls`
>       
>       * make connection against `url`
>       * if response is success, break from the loop
> 
>  * outside of the loop:
>       
>       * if everything failed -> raise an error ; else process the successful byte stream

That's exactly what the code does.

src/client.rs
seanyoung: I don't think you understood the code.

src/client.rs
flavio: Sorry, I'm dumb :sweat: , I missed check on line `878`.

So to recap, we try to get the blob by using the url built from the digest of the layer first. If that fails, then we iterate over the urls until we find a working one. Is that right?

I think this could be made more explicit with something like:

```rust
let url = self.to_v2_blob_url(image.resolve_registry(), image.repository(), &layer.digest);

let mut response = RequestBuilderWrapper::from_client(self, |client| client.get(&url))
    .apply_accept(MIME_TYPES_DISTRIBUTION_MANIFEST)?
    .apply_auth(image, RegistryOperation::Pull)?
    .into_request_builder()
    .send()
.await?;

if response.error_for_status_ref().is_err() && let Some(urls) = &layer.urls {
    for url in urls {
        let url = Url::parse(url)
            .map_err(|e| OciDistributionError::UrlParseError(e.to_string()))?;

        if url.scheme() == "http" || url.scheme() == "https" {
            // NOTE: we must not authenticate on additional URLs as those
            // can be abused to leak credentials or tokens.  Please
            // refer to CVE-2020-15157 for more information.
            response =
                RequestBuilderWrapper::from_client(self, |client| client.get(url.clone()))
                    .apply_accept(MIME_TYPES_DISTRIBUTION_MANIFEST)?
                .into_request_builder()
                .send()
            .await?;

            if response.error_for_status_ref().is_ok() {
                break;
            }
        }
    }
}
```

However, that's a matter of personal taste. I wonder what @thomastaylor312 thinks about that

src/client.rs
flavio: Yes, my fault. I'm sorry

src/client.rs
seanyoung: same as `fn linux_amd64_resolver` but for windows containers

src/client.rs
thomastaylor312: I am ok with either way at this point. So unless you feel strongly about having it this way for readability reasons, we can keep as is

thomastaylor312:
Hmmm really odd test failure:

```
thread 'client::test::test_mount' panicked at src/client.rs:2671:14:
Failed to mount: SpecViolationError("Expected HTTP Status 201 Created, got 202 Accepted instead")
```

@seanyoung you seeing this locally as well? Gonna pull down this branch and try

thomastaylor312:
Weirdly enough, it does look like the spec says it should be returning 201:  https://github.com/opencontainers/distribution-spec/blob/main/spec.md#endpoints. Tests work on main though, so I'm looking into this

src/client.rs
```diff
@@ -2591,8 +2653,12 @@ mod test {
             .parse()
             .unwrap();
         let layer_data = vec![1u8, 2, 3, 4];
-        let layer_digest = sha256_digest(&layer_data);
-        c.push_blob(&layer_reference, &[1, 2, 3, 4], &layer_digest)
+        let layer = OciDescriptor {
```
thomastaylor312, line 2656: Ok, figured out why this was failing. The value returned from `push_blob` is the full URL, not just the sha:

```
layer digest: http://localhost:55038/v2/layer-repository/blobs/sha256:9f64a747e1b97f131fabb6b447296c9b6f0201e79fb3c5356e6c77e89b6a806a
layer digest struct: sha256:9f64a747e1b97f131fabb6b447296c9b6f0201e79fb3c5356e6c77e89b6a806a
thread 'client::test::test_mount' panicked at src/client.rs:2674:14:
Failed to mount: SpecViolationError("Expected HTTP Status 201 Created, got 202 Accepted instead")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

So in the `mount_blob` call beneath this, you'll need to do `&layer.digest` instead

seanyoung: That makes sense, it's been fixed. Thanks for finding that

src/client.rs
flavio: let's keep it as it currently is :+1:

flavio (approved):
All the tests are green. I think we're good to go

src/client.rs
flavio: It's solved now

flavio:
@thomastaylor312 I would go ahead and merge this PR if you're fine with that

thomastaylor312 (approved):
Thanks again @seanyoung!

PR #109, opened 2024-01-28
client is now clonable and can be used in concurrent manner

aviramha:
Hey all!
Thanks for the great crate.
I'm starting a new project that relies heavily on it, and noticed that `Client` isn't `Clone` for no apparent reason - adding it is easy and enables users to concurrently work with it ;)

PR #110, opened 2024-01-28
Add streaming push

aviramha:
New function - `push_blob_stream` lets users pass a stream to the blob upload function in order for it to iterate.
The signature is similar to the one of the pull so those could be chained.
Builds upon #109


aviramha:
I'm not sure about this PR because when I tested it it took infinite time to run opposed to the "one off" functions. I'll finish the re-auth first.

flavio:
@aviramha can you rebase this PR against main?

Also, @thomastaylor312 and I would like to tag a new release of this crate once this PR gets merged :)

aviramha:
I think it'll take me some time to re-visit, while I tried to use this flow I couldn't manage to get it to work so I fear there needs to be more thinking on the design.

flavio:
@aviramha no worries, I don't think we're on a hurry to push out a new version of the crate

PR #111, opened 2024-01-29
Refactor auth - add re-auth mechanism

aviramha:
Sorry about the spam!
This builds upon the other 2 PRs #109  #110 
Just out of comfort assuming if merged I'll work on it serially.
This adds logic to enable token re-creation when needed, and also making the auth API more fluent.
I had a situation where my token got expired mid-upload and there was no place to renew, and if I could break the API I'd have structured it a bit different, but I think is ok.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -534,6 +534,8 @@ impl Client {
         authentication: &RegistryAuth,
         operation: RegistryOperation,
     ) -> Result<Option<String>> {
+        self.store_auth_if_needed(image.resolve_registry(), authentication)
+            .await;
         // preserve old caching behavior
         match self._auth(image, authentication, operation).await {
             Ok(Some(RegistryTokenType::Bearer(token))) => {
@@ -1801,6 +1803,14 @@ mod test {
             .sign_with_key(&key)?
             .as_str()
             .to_string();
+
+        // we have to have it in the stored auth so we'll get to the token cache check.
+        client
+            .store_auth(
+                &Reference::try_from(HELLO_IMAGE_TAG)?.resolve_registry(),
+                RegistryAuth::Anonymous,
+            )
+            .await;
 
         client
             .tokens
--- a/src/client.rs
+++ b/src/client.rs
@@ -33,6 +33,7 @@ use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::sync::Arc;
 use tokio::io::{AsyncWrite, AsyncWriteExt};
+use tokio::sync::RwLock;
 use tracing::{debug, trace, warn};
 
 const MIME_TYPES_DISTRIBUTION_MANIFEST: &[&str] = &[
@@ -205,6 +206,8 @@ impl TryFrom<Config> for ConfigFile {
 #[derive(Clone)]
 pub struct Client {
     config: Arc<ClientConfig>,
+    // Registry -> RegistryAuth
+    auth_store: Arc<RwLock<HashMap<String, RegistryAuth>>>,
     tokens: TokenCache,
     client: reqwest::Client,
     push_chunk_size: usize,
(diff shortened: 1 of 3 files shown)


src/token_cache.rs
```diff
@@ -59,7 +59,15 @@ pub enum RegistryOperation {
     Pull,
 }
 
-type CacheType = BTreeMap<(String, String, RegistryOperation), (RegistryTokenType, u64)>;
+// Types to allow better naming
+type Registry = String;
+type Repository = String;
+type TokenCacheKey = (Registry, Repository, RegistryOperation);
+type TokenExpiration = u64;
+type TokenCacheValue = (RegistryTokenType, TokenExpiration);
```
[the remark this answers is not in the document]
aviramha: Done.

flavio:
@aviramha can you please look into the unit tests that are failing? I can reproduce the failures locally, and they are a regression introduced by this PR

aviramha:
Sure! I had an assumption those tests require auth. Fixed now.

src/client.rs
```diff
@@ -502,6 +534,38 @@ impl Client {
         authentication: &RegistryAuth,
         operation: RegistryOperation,
     ) -> Result<Option<String>> {
+        self.store_auth_if_needed(image.resolve_registry(), authentication)
+            .await;
+        // preserve old caching behavior
```
thomastaylor312, line 539: Do we actually need to preserve the old behavior? I think a call to `get_auth_token` should do the same logic of fetching the token and doing the auth step, right?

src/token_cache.rs
```diff
@@ -59,7 +59,26 @@ pub enum RegistryOperation {
     Pull,
 }
 
-type CacheType = BTreeMap<(String, String, RegistryOperation), (RegistryTokenType, u64)>;
+// Types to allow better naming
+type Registry = String;
```
thomastaylor312, line 63: Nit: I don't really like aliased types like this because they lead to confusion down the line (an extra mental hop to figure out it is a string). I can get behind them if defining a complex type, but if they are just used in a struct like they are below, it is pretty clear what it is for due to the field name

src/token_cache.rs
```diff
@@ -132,34 +149,33 @@ impl TokenCache {
...
+                expiration,
+            }) => {
                 let now = SystemTime::now();
                 let epoch = now
                     .duration_since(UNIX_EPOCH)
                     .expect("Time went backwards")
                     .as_secs();
                 if epoch > *expiration {
-                    debug!(%registry, %repository, ?op, %expiration, miss=false, expired=true, "Fetching token");
+                    debug!(?key, %expiration, miss=false, expired=true, "Fetching token");
```
thomastaylor312, line 168: I think we should still log each of the different fields separately here. The default debug implementation might be really "noisy" when trying to look at these logs and it will be hard to filter them based on operation or repo in any of the logging tools

thomastaylor312 (changes requested):
Just a few minor changes and discussion points. Nits are optional to address! Thanks for implementing this!

src/client.rs
aviramha: I didn't want to break API, and that could be that user wants to do auth and it might not really call auth underneath (because cache) feels a bit vile.

src/token_cache.rs
flavio: solved

src/token_cache.rs
flavio: fixed

src/client.rs
thomastaylor312: Ok, that makes sense

thomastaylor312 (approved):
I think we're good to go! @flavio leaving merging to you since you started reviewing/conversation on this one

PR #115, opened 2024-02-10
Implement Client::pull_manifest_raw

### LINKED ISSUE #114: Retrieving "unmangled" manifests for digest consistency?
Is there a way to fetch raw manifests without deserializing them immediately?

When retrieving manifests, they're automatically deserialized into Rust structures which is very handy, but unfortunately this also means that if you intend to push the manifest elsewhere your re-serialized version of the content might not match the digest provided by the upstream.

~~This problem can be worked around by simply not providing an expected digest when pushing to containerd for example, but this feels like a bit of a hack, since the changes to the underlying manifest (if any) are purely cosmetic, and the content and digest are in fact inconsistent.~~ Removing the expected digest simply allows the push to go through, but the final object will still have a different digest.

MathiasPius:
Implements `Client::pull_manifest_raw` for fetching the plain text representation of a manifest without deserializing it into an `OciManifest`.

This way the manifest can be pushed elsewhere while maintaining the digest value of the manifest which is otherwise subject to change following re-serialization of the `OciManifest` object.

### Asymmetry with `push_raw_manifest`
`push_manifest_raw` takes a `Vec<u8>`, while `pull_manifest_raw` returns the content as a `String`. I've chosen this route, since it allows us to still call `validate_registry_response` in `Client::_pull_manifest_raw` where the response status and manifest text is available for validation.

### Providing `accepted_media_types` when pulling raw manifests
Since `pull_manifest_raw` is specifically intended for fetching individual manifests for mirroring, I expect the caller to be traversing the tree of manifests manually and pushing/storing them elsewhere, and in that case they will have a very good idea of what kind of manifest they're likely to get back at each individual step.

### Deduplicating code
I've re-implemented `Client::_pull_manifest` in terms of `Client::_pull_manifest_raw` to reduce code duplication, so it now uses the raw function to get the text manifest, and then performs deserialization from there.

Closes #114

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -364,7 +364,7 @@ impl Client {
 
         validate_registry_response(status, &body, &url)?;
 
-        Ok(serde_json::from_str(&std::str::from_utf8(&body)?)?)
+        Ok(serde_json::from_str(std::str::from_utf8(&body)?)?)
     }
 
     /// Pull an image and return the bytes
@@ -856,10 +856,10 @@ impl Client {
 
         let text = std::str::from_utf8(&body)?;
 
-        self.validate_image_manifest(&text).await?;
+        self.validate_image_manifest(text).await?;
 
         debug!("Parsing response as Manifest: {}", &text);
-        let manifest = serde_json::from_str(&text)
+        let manifest = serde_json::from_str(text)
             .map_err(|e| OciDistributionError::ManifestParsingError(e.to_string()))?;
         Ok((manifest, digest))
     }
--- a/src/client.rs
+++ b/src/client.rs
@@ -679,7 +679,7 @@ impl Client {
             let text = res.text().await?;
             validate_registry_response(status, &text, &url)?;
 
-            digest_header_value(headers, Some(&text))
+            digest_header_value(headers, Some(&text.as_bytes()))
         } else {
             let status = res.status();
             let headers = res.headers().clone();
@@ -840,7 +840,7 @@ impl Client {
 
         validate_registry_response(status, &text, &url)?;
 
-        let digest = digest_header_value(headers, Some(&text))?;
+        let digest = digest_header_value(headers, Some(&text.as_bytes()))?;
 
         Ok((text, digest))
     }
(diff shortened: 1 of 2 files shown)


src/client.rs
```diff
@@ -731,6 +731,25 @@ impl Client {
...
+    /// not will attempt to do.
+    ///
+    /// A Tuple is returned containing the plain text representation of the manifest
+    /// and the manifest content digest hash.
+    pub async fn pull_manifest_raw(
+        &self,
+        image: &Reference,
+        auth: &RegistryAuth,
+        accepted_media_types: &[&str],
+    ) -> Result<(String, String)> {
```
thomastaylor312, line 746: I still think I'd prefer to have the symmetry of this being a `Vec<u8>` because I think that is what would be expected when pulling something raw. I totally understand the motivation behind keeping it a string though. I think the validate function should just be updated to take `&[u8]`, but I don't want to make you do that if you don't want to. So if you'd like to do it, let me know. Otherwise, I can merge this after a day or so if I haven't heard from you and make the change in a follow up PR!

thomastaylor312 (approved):
Waiting for a day to merge per my review comment. Thanks for adding this @MathiasPius!

src/client.rs
flavio: I agree with @thomastaylor312 on this request

src/client.rs
```diff
@@ -2717,6 +2752,43 @@ mod test {
         assert_eq!(manifest.config.digest, pulled_manifest.config.digest);
     }
 
+    #[tokio::test]
+    #[cfg(feature = "test-registry")]
+    async fn test_raw_manifest_digest() {
```
flavio, line 2757: This test is not making any request against the local registry. You can drop the feature `test-restry` and the code that deal with starting the local registry

flavio (changes requested):

src/client.rs
MathiasPius: Ah correct! In a previous iteration I did have it use the local registry, but it didn't really add any assurances to the test.

src/client.rs
MathiasPius: Perfectly reasonable, I'll see what I can do!

MathiasPius:
Thanks for the quick reviews! I've implemented the changes as requested, please review the updated PR, and let me know if there's anything else :)

flavio:
@MathiasPius can you please fix the small formatting issue that causes the windows test to fail?

Aside from that, the PR LGTM

flavio (approved):
Thanks for having addresses all the issues

PR #121, opened 2024-03-29
ref(client): Debug, Eq, Hash, PartialEq for ImageLayer; convert annotations to BTreeMap

vdice:
- Update ImageLayer to derive Debug, Eq, Hash and PartialEq traits
- **_Breaking_**: Convert annotations from a HashMap to a BTreeMap per https://github.com/krustlet/oci-distribution/pull/121#discussion_r1544777159, so that we don't need a custom Hash impl for ImageLayer

Changeset motivated by the need to differentiate and deduplicate collections of ImageLayers.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -23,16 +23,15 @@ use futures_util::stream::{self, StreamExt, TryStreamExt};
 use futures_util::Stream;
 use http::HeaderValue;
 use http_auth::{parser::ChallengeParser, ChallengeRef};
-use itertools::Itertools;
 use olpc_cjson::CanonicalFormatter;
 use reqwest::header::HeaderMap;
 use reqwest::{RequestBuilder, Url};
 use serde::Deserialize;
 use serde::Serialize;
 use sha2::Digest;
-use std::collections::HashMap;
+use std::collections::{BTreeMap, HashMap};
 use std::convert::TryFrom;
-use std::hash::{Hash, Hasher};
+use std::hash::Hash;
 use std::sync::Arc;
 use tokio::io::{AsyncWrite, AsyncWriteExt};
 use tokio::sync::RwLock;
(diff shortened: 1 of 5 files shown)


src/client.rs
```diff
@@ -125,6 +127,22 @@ impl ImageLayer {
     }
 }
 
+/// Implementing Hash to enable determining uniqueness
+/// eg via https://docs.rs/itertools/latest/itertools/structs/struct.Unique.html
+impl Hash for ImageLayer {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.data.hash(state);
+        self.media_type.hash(state);
+        if let Some(annotations) = self.annotations.clone() {
```
thomastaylor312, line 136: I think I'd rather swap annotations to be a `BTreeMap` and break the API a little bit. It makes sense that we'd want these things to be hashable and comparable and the `BTreeMap` is already `Hash` and `Eq`. That way we don't need another dependency and the traits can be entirely derived

thomastaylor312 (changes requested):
Great change here! One comment around changing the type and just breaking the API

src/client.rs
vdice: 👍 Nice, much simpler.  I left the unit test in...

thomastaylor312 (approved):
Thanks for the great work here! @flavio would like your review too before I merge since this is a breaking API change.

src/client.rs
thomastaylor312: Yeah I'm totally fine with itertools as a dev dep

thomastaylor312:
@vdice heads up that it looks like the BTreeMap import didn't get picked up in the windows build for some reason

vdice:
@thomastaylor312 thanks - looks like I didn't update the import in examples/wasm/push.rs.  Now updated; hopefully that was the issue.

thomastaylor312:
Thanks for soldering through that @vdice. Just waiting on a second opinion from @flavio on the breaking aspect of this

flavio:
Sorry, I was on vacation. This change looks good to me

vdice:
Thanks so much @thomastaylor312 and @flavio!

PR #124, opened 2024-04-07
Create new LayerDescriptor for pulling blobs

SuperTux88:
The old `OciDescriptor` has too many fields, which maybe aren't available when trying to pull a blob with only the digest. In the past you only needed the digest as string, but the optional urls field was added (in #108). But all other fields of the `OciDescriptor` aren't required when pulling a blob, but are required by the struct. The `LayerDescriptor` struct now has only the fields which are used for pulling.

This is a breaking change, as it changes the signature of the pull methods. But with this it is now possible to pull a blob with only the digest as a `&str` again.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -94,6 +94,15 @@ pub struct LayerDescriptor<'a> {
     pub urls: &'a Option<Vec<String>>,
 }
 
+impl<'a> From<&'a LayerDescriptor<'a>> for LayerDescriptor<'a> {
+    fn from(descriptor: &'a LayerDescriptor<'a>) -> Self {
+        LayerDescriptor {
+            digest: descriptor.digest,
+            urls: descriptor.urls,
+        }
+    }
+}
+
 impl<'a> From<&'a str> for LayerDescriptor<'a> {
     fn from(digest: &'a str) -> Self {
         LayerDescriptor {
@@ -109,6 +118,45 @@ impl<'a> From<&'a OciDescriptor> for LayerDescriptor<'a> {
             digest: &descriptor.digest,
             urls: &descriptor.urls,
         }
+    }
+}
+
+/// A trait for converting any type into a [`LayerDescriptor`]
+pub trait AsLayerDescriptor {
+    /// Convert the type to a LayerDescriptor reference
+    fn as_layer_descriptor(&self) -> LayerDescriptor<'_>;
+}
+
+impl AsLayerDescriptor for &str {
+    fn as_layer_descriptor(&self) -> LayerDescriptor<'_> {
+        LayerDescriptor {
+            digest: self,
+            urls: &None,
+        }
+    }
+}
+
+impl AsLayerDescriptor for OciDescriptor {
+    fn as_layer_descriptor(&self) -> LayerDescriptor<'_> {
+        LayerDescriptor {
+            digest: &self.digest,
+            urls: &self.urls,
+        }
+    }
+}
+
+impl<'a> AsLayerDescriptor for &'a OciDescriptor {
+    fn as_layer_descriptor(&self) -> LayerDescriptor<'a> {
+        LayerDescriptor {
+            digest: &self.digest,
+            urls: &self.urls,
+        }
+    }
+}
+
+impl AsLayerDescriptor for LayerDescriptor<'_> {
+    fn as_layer_descriptor(&self) -> LayerDescriptor<'_> {
+        self.into()
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/client.rs
```diff
@@ -965,10 +992,10 @@ impl Client {
     pub async fn pull_blob<T: AsyncWrite + Unpin>(
         &self,
         image: &Reference,
-        layer: &OciDescriptor,
+        layer: &LayerDescriptor<'_>,
```
thomastaylor312, line 995: Since `LayerDescriptor` contains purely borrowed types, I think I'd prefer this not also be a borrowed parameter. That would avoid a call that looks like this: `&(&layer).into()` like we have down below

src/client.rs
```diff
@@ -86,6 +86,32 @@ pub struct TagResponse {
...
+impl<'a> From<&'a str> for LayerDescriptor<'a> {
+    fn from(digest: &'a str) -> Self {
+        LayerDescriptor {
+            digest,
+            urls: &None,
+        }
+    }
+}
+
+impl<'a> From<&'a OciDescriptor> for LayerDescriptor<'a> {
```
thomastaylor312, line 106: I like having this implementation too, but I think it would be a less painful breaking change and a cleaner API if `OciDescriptor` implemented `Deref` for `LayerDescriptor` if we can. That means a call like `c.pull_blob(&image_reference, &(&layer).into(), &mut buf)` would be `c.pull_blob(&image_reference, &layer, &mut buf)`

thomastaylor312 (changes requested):
Love this change. Just have some additional feedback around the API design

src/client.rs
SuperTux88: I would also prefer to just have `&layer` instead of `&(&layer).into()`, if that's possible, but I can't get it to work. But I'm also still pretty new to rust and haven't used `Deref` yet, so maybe I'm doing something wrong? But my problem is, that the `Target` type (`LayerDescriptor`) requires a lifetime parameter, but `OciDescriptor` doesn't have one, so I get `error[E0207]: the lifetime parameter `'a` is not constrained by the impl trait, self type, or predicates`.

I'd be happy to receive some hint what I'm doing wrong, but since you wrote "if we can", I'm not sure if you know if it's possible or not?

thomastaylor312: Let me give it a whirl real quick. Are you ok if I push a commit to this branch?

src/client.rs
thomastaylor312: Yep, I wondered if that was the case. Let's see if we can figure out the `Deref` thing and then make a call from there

src/client.rs
thomastaylor312: Yep, you were right around using `Deref`, so I used another way of doing this I've seen before and pushed it up to my own branch. I didn't want to push to yours without your sign off, but you are also welcome to just take this code if you'd like a push it. Let me know what you think about it: https://github.com/krustlet/oci-distribution/commit/0a1443eb0ba46d183ceeb48c106150d5c84ab51d

SuperTux88: > Let me give it a whirl real quick. Are you ok if I push a commit to this branch?

Sure, there is this checkbox, where I can enable or block you from being able to push to my branch, and I intentionally always leave that enabled, because I think there are many usecases where maintainers are able to do that, from just fixing a small last nitpick before merging (instead of waiting for the original author to do it themselves) to suggestions and collaboration like here.
image
Also, git is very flexible and you can always undo and bring back an earlier state or do something completely different later, so even if I wouldn't like your commit, I could remove it again.

> so I used another way of doing this I've seen before

Great, that works, and I like how the results of using the API looks with this :+1: So I pulled in your commit, and also changed it for the `pull_blob_stream()` to work the same.

I then added another impl for `AsLayerDescriptor` to work with borrowed instances of `LayerDescriptor`. This solves my use-case from the other comment, where I also wanted to use instances of `LayerDescriptor` in my own code, and I can then also just use that as borrowed parameter for pulling.

What I didn't like about the new solution was, how much code and complexity it added with all the different impls. It added a lot of code which all would be needed to be maintained in the future, so I removed all the impls again, which I think aren't needed (they can still be easily added back if a use-case for them comes up). What remains are:
* `impl AsLayerDescriptor for &str` for when you only have a digest as string.
* `impl AsLayerDescriptor for &OciDescriptor` for when you have a full `OciDescriptor`, like it's used internally, this can then be borrowed for pulling
* `impl AsLayerDescriptor for &LayerDescriptor<'_>` which I added for when you already work with an instance of `LayerDescriptor` and want to borrow that.

So you either have one of the first two, which can be used to convert to `LayerDescriptor` if needed, and then work with that and use the third.

thomastaylor312: > Also, git is very flexible and you can always undo and bring back an earlier state or do something completely different later, so even if I wouldn't like your commit, I could remove it again.

Of course! I still just like asking to be polite!

Taking a look now

thomastaylor312 (approved):
Thanks for the work here!

PR #141, opened 2024-06-25
ref(*): Updates the repo to reflect its new location

thomastaylor312:
This does some cleanup of the readme as well to remove old/stale info and updates the repo in Cargo.toml


shizhMSFT (approved):
LGTM

flavio:
Should we go ahead and rename the crate too?

thomastaylor312:
@flavio Let's open an issue and tag everyone we know who is using it. I don't think there will be any complaints, but it will also serve as an informational issue so people know we'll be pushing it to a new `oci-client` crate

PR #143, opened 2024-06-28
feat(client): allow to set the User-Agent used by the client

SuperTux88:
I noticed that the client doesn't send a user-agent, so lets set the crate name/version by default but also allow others to overwrite it with their own app-name in the config.


thomastaylor312 (approved):
Thanks for adding this!

PR #145, opened 2024-07-05
Partial Requests and some other pull refactorings

SuperTux88:
The main reason for this Pull-Request is to allow doing partial requests. This is a very useful feature if you have very slow/unstable internet (like in Germany) and the pull often fails for various reasons (timeout or other connection errors). So it doesn't help if you can download the first few MB of a blob over and over again before the network connection craps out after some minutes and you can start over again. If your upstream registry supports partial requests, you at least can pull the next few MB every time, and eventually get the whole blob.

I only added the streaming version of this function, as I didn't know how to handle it when giving a writer as parameter, as if the upstream registry doesn't support it, it will respond with a full response and there is no good way to signal that (except maybe just return an error if the upstream registry responded with `200 OK`, but I didn't really like this idea). But let me know if you want me to add something. For my use-case the streaming version is enough and I can signal if it's a full or partial response using an enum.

I also did some refactoring to re-use some mostly identical code instead of duplicating it.

And then I also changed the return type of the streaming variants to also include the `Content-Length` header of the response (if available), so you can use that information without reading the whole response (for example to also set that header when proxying). This is a breaking change for the `pull_blob_stream(...)` function, as it changes the return type, but I hope that's fine as in #124 was already a breaking change for that function, which isn't released yet.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -25,7 +25,7 @@ use http::HeaderValue;
 use http_auth::{parser::ChallengeParser, ChallengeRef};
 use olpc_cjson::CanonicalFormatter;
 use reqwest::header::HeaderMap;
-use reqwest::{RequestBuilder, Url};
+use reqwest::{RequestBuilder, Response, Url};
 use serde::Deserialize;
 use serde::Serialize;
 use sha2::Digest;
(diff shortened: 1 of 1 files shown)


thomastaylor312:
@SuperTux88 Sorry for the delay here! Got pulled away to work on a few other things. I'll get to this tomorrow!

src/client.rs
```diff
@@ -1097,22 +1124,35 @@ impl Client {
...
+                    response = request.send().await?
                 }
             }
         }
 
+        Ok(response)
+    }
+
+    /// Converts a response into a stream
+    fn stream_from_response(&self, response: Response) -> Result<SizedStream> {
```
thomastaylor312, line 1146: Nit: This can probably be a helper function that lives outside of the `Client` impl since it doesn't depend on anything in `self`

src/client.rs
```diff
@@ -173,6 +173,22 @@ impl ImageLayer {
...
+/// Stream response of a blob with optional content length if available
+pub struct SizedStream {
+    /// The length of the stream if the upstream registry sent a `Content-Length` header
+    pub content_length: Option<u64>,
+    /// The stream of bytes
+    pub stream: BoxStream<'static, std::result::Result<bytes::Bytes, std::io::Error>>,
+}
+
+/// The response of a partial blob request
+pub enum PartialBlobResponse {
```
thomastaylor312, line 185: I think this should probably just be called `BlobResponse` because it reads a little weird to have `PartialBlobResponse::Partial` (repetitive) and `PartialBlobResponse::Full` (contradictory). I know it is technically only being returned from the blob partial method, so I could definitely be argued out of it.

thomastaylor312 (commented):
Great job on the refactor here and the new feature. One minor comment around naming of one of the structs and a nit

src/client.rs
SuperTux88: Yeah, that are probably old habits from other programming languages, like Java for example, where just everything is in the scope of a class, so I just didn't think about that I could move this helper function outside and remove `&self`, so thanks for noticing :smile: 

(the only advantage it had with being inside of `Client` was, that it was close to where it was used, so it's maybe easier to read the code when you don't need to jump around that much, but with modern IDEs that's probably not a big problem)

src/client.rs
SuperTux88: Yeah, I was unsure about that. Initially I had it named `BlobResponse`, but then it felt weird, because with that name you would kinda expect that the `pull_blob_stream(...)` function would always return a `BlobResponse::Full`, which wouldn't make much sense to always wrap that response if `Full` is the only variant you get from that method? :man_shrugging: So I ended up renaming it. But you are right, the `PartialBlobResponse` name also is weird.

But since I was unsure which name is better anyway, I'm also happy to just rename it back. :+1:

thomastaylor312 (approved):

PR #148, opened 2024-07-24
Add namespace query parameter for registry mirroring/proxying

SuperTux88:
This adds the ability to set a mirror to the `Reference` instance, which is then used to pull the image, instead of the registry that was parsed from the image name. And it then forwards the original registry with the `ns` query parameter to the mirror registry, so the mirror registry can use this information to decide from which registry you wanted to pull something.

This is currently not an officially documented standard feature yet, but it's probably going to be eventually, it's just taking a really long time (https://github.com/opencontainers/distribution-spec/pull/66). But containerd is already using the `ns` query parameter and I think I read somewhere that nexus is also already using the parameter when used as pull through cache (but I haven't verified that). But I think it's a really useful feature (when working with mirrors/proxies), even if it's not officially part of the spec yet, so I also implemented it to the library here.

Some additional notes:
* It currently doesn't really handling pushing after you set the mirror registry (as it would also try to push through the mirror, which probably wouldn't work). But my assumption was, that somebody who set the mirror registry doesn't also use the same `Reference` instance also for pushing the same image again (would that even make sense if you just pulled it). So I decided to not add additional complexity for that use-case in the code for now. For people who don't set a mirror registry, everything works the same as before and pushing works. And if it's really needed, this logic can always be added later, but I wanted to keep the code simpler. But I'm open to feedback.
* I wasn't sure if I should make the parameter for the `set_mirror_registry()` function an `Option<String>`, so you can also unset the mirror again, but for my use-case I don't need it to be unset. But let me know what you think and I can change it.
* I have no idea why `cargo fmt` formats the `rstest` at the `test_mirror_registry` like that, but doesn't do it for the other tests :man_shrugging: 

Also: The last commit doesn't have anything to do with this feature, but when scrolling through the code while adding this feature, I saw the `FIXME` and thought "why not just fix that as well?" :smiley:

### DIFF
--- a/src/reference.rs
+++ b/src/reference.rs
@@ -104,6 +104,17 @@ impl Reference {
         }
     }
 
+    /// Clone the Reference for the same image with a new digest.
+    pub fn clone_with_digest(&self, digest: String) -> Self {
+        Self {
+            registry: self.registry.clone(),
+            mirror_registry: self.mirror_registry.clone(),
+            repository: self.repository.clone(),
+            tag: None,
+            digest: Some(digest),
+        }
+    }
+
     /// Set a pull mirror registry for this reference.
     ///
     /// The mirror registry will be used to resolve the image, the original registry
--- a/src/reference.rs
+++ b/src/reference.rs
@@ -256,33 +256,25 @@ impl TryFrom<String> for Reference {
         }
         // Digests much always be hex-encoded, ensuring that their hex portion will always be
         // size*2
-        if reference.digest().is_some() {
-            let d = reference.digest().unwrap();
-            // FIXME: we should actually separate the algorithm from the digest
-            // using regular expressions. This won't hold up if we support an
-            // algorithm more or less than 6 characters like sha1024.
-            if d.len() < 8 {
-                return Err(ParseError::DigestInvalidFormat);
-            }
-            let algo = &d[0..6];
-            let digest = &d[7..];
-            match algo {
-                "sha256" => {
+        if let Some(digest) = reference.digest() {
+            match digest.split_once(':') {
+                None => return Err(ParseError::DigestInvalidFormat),
+                Some(("sha256", digest)) => {
                     if digest.len() != 64 {
                         return Err(ParseError::DigestInvalidLength);
                     }
                 }
-                "sha384" => {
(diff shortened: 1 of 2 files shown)


thomastaylor312:
> I saw the FIXME and thought "why not just fix that as well?"

❤️

src/reference.rs
```diff
@@ -133,6 +157,15 @@ impl Reference {
         self.digest.as_deref()
     }
 
+    /// ns returns the original registry when pulled via a mirror.
+    pub fn ns(&self) -> Option<&str> {
```
thomastaylor312, line 161: Can this just be called `namespace` to match the proposal? I know it maps to the parameter, but it is a bit confusing when reading

src/reference.rs
```diff
@@ -95,21 +97,43 @@ impl Reference {
...
+            tag: None,
+            digest: Some(digest),
+        }
+    }
+
+    /// Set a pull mirror registry for this reference.
+    ///
+    /// The mirror registry will be used to resolve the image, the original registry
+    /// is available via the [`Reference::ns`] function.
+    pub fn set_mirror_registry(&mut self, registry: String) {
```
thomastaylor312, line 122: We should probably have this and the `ns` function be `#[doc(hidden)]` with information in their doc strings about this not being stable yet (and as such, is exempt from semver backwards compat guarantees). I am perfectly fine with it being in here if things like containerd are already using it, but I want to make it clear that this isn't fully supported yet. The other option would be to have something like a `ClientExp` trait that the client could implement for experimental features, but that also feels a bit extra. 

Like you mentioned, it looks like it will land, but things can take a while in the OCI spec, so I want to be extra careful here

thomastaylor312 (changes requested):
Generally in favor of having this, left some comments with my concerns.

Thanks again for cleaning up the FIXME!

src/reference.rs
SuperTux88: Sure, I added the `#[doc(hidden)]` and added a note about it not being stable yet, and also added a note that the `ns` query parameter might not work with all registries yet.

thomastaylor312 (approved):

PR #150, opened 2024-07-29
feat(client): Make store_auth_if_needed public

SuperTux88:
This allows to store auth information to the client if needed, without doing a full auth request.

I have the use-case where I might pull a blob without pulling the manifest before, as there are multiple replicas and the manifest might have been pulled by another instance already. But that means that some instances never had the auth information stored and therefore the pull of the blob fails with a 401.

I know I could all `client.auth(...)`, but that also runs a full auth request (which even is multiple requests) every time, even if the auth info is already stored for this registry, or even when it's already authenticated for this image.

I also could keep track myself if I already authenticated for a certain registry, but that would be kinda redundant if the client already keeps track of it.

And I understand that you don't want to require to provide auth information as required parameter when pulling a blob, as if you do everything with one instance, it should already know it from pulling the manifest. So I think the best solution is just to make the `store_auth_if_needed(...)` public, so I can do exactly what I need to do, give the auth info to the client so it can use it if required for pulling a blob, without affecting any other use-cases and workflows.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -374,7 +374,15 @@ impl Client {
         self.auth_store.read().await.contains_key(registry)
     }
 
-    async fn store_auth_if_needed(&self, registry: &str, auth: &RegistryAuth) {
+    /// Store the authentication information for this registry if it's not already stored in the client.
+    ///
+    /// Most of the time, you don't need to call this method directly. It's called by other
+    /// methods (where you have to provide the authentication information as parameter).
+    ///
+    /// But if you want to pull/push a blob without calling any of the other methods first, which would
+    /// store the authentication information, you can call this method to store the authentication
+    /// information manually.
+    pub async fn store_auth_if_needed(&self, registry: &str, auth: &RegistryAuth) {
         if !self.is_stored_auth(registry).await {
             self.store_auth(registry, auth.clone()).await;
         }


src/client.rs
```diff
@@ -374,7 +374,8 @@ impl Client {
         self.auth_store.read().await.contains_key(registry)
     }
 
-    async fn store_auth_if_needed(&self, registry: &str, auth: &RegistryAuth) {
+    /// Store the authentication information for this registry if it's not already stored in the client
```
thomastaylor312, line 377: Can you add a bit more about when you'd use this? You could probably copy/paste most of it from the PR description

src/client.rs
```diff
@@ -374,7 +374,8 @@ impl Client {
...
+    pub async fn store_auth_if_needed(&self, registry: &str, auth: &RegistryAuth) {
```
thomastaylor312, line 378: This is a nit/ask for your opinion, but what do you think about calling this `store_auth` instead and then documenting that it only stores it if needed? `store_auth_if_needed` seems like a weird public function name to me 😅

thomastaylor312 (commented):
Super minor requests. I think I am good with exposing this publicly

src/client.rs
SuperTux88: Sure, I added a bit more info about when you need this. :+1:

src/client.rs
SuperTux88: I'm not sure about this one. There is already a `store_auth` function (it gets called by this one), and I was even thinking about making that one public as well. Because that could also be handy, as there is currently no way to overwrite the auth info if it's once set. And it gets set before it actually checks if the provided credentials even work. Or you might have rolling credentials and need to rotate them. And if you rename this one to `store_auth`, then you need another name for the function again which always overwrites the auth info.

I don't need this use-cases yet (so I only made `store_auth_if_needed` public for now), but might have them in the future (and it's maybe also useful for others to be able to change credentials). So what are your thoughts about also making `store_auth` public, and I think then the current name with the `_if_needed` makes sense if both of them are available publicly.

thomastaylor312: Since the use case you have for `store_auth_if_needed`, let's just make that public as is. With the additional documentation, it should be clear why it is named that way

thomastaylor312 (approved):

PR #151, opened 2024-07-29
ref(*)!: Renames the crate to oci-client

### LINKED ISSUE #142: Rename crate to `oci-client`
We've recently moved this code from `krustlet/oci-distribution` over there (`oras-project/rust-oci-client`).

Should we also rename the crate from `oci-distribution` to `oci-client`?

We could then make PRs against the projects that are using `oci-distribution` to have them moved to the new crate. An approximate list can be found [here](https://github.com/search?q=oci-distribution+language%3ATOML&type=code&l=TOML)

Reading on the internet, we could issue an informal advisory inside of RustSec. In this way, `cargo-audit` would inform all the end users of the crate about the rename

thomastaylor312:
Closes #142


thomastaylor312:
Once this is merged, I'll push (locally) a modified version of 0.11 to crates.io from the 0.11.0 tag with the new name. Once that is there I can open the rust advisory.

flavio (approved):
thanks a lot for having taken care of that

PR #152, opened 2024-07-30
fix(client): Validate digests when header is set

### LINKED ISSUE #149: image pulling does not validate digests
When pulling an image by digest, I would expect the library to verify that the digest of the returned resource matches the requested digest.

For testing purposes, I set up a registry that returns the same image for any requested digest:

```
$ curl -s https://evil-registry.rudy.family/v2/busybox/manifests/sha256:deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef | sha256sum
d319b0e3e1745e504544e931cde012fc5470eba649acc8a7b3607402942e5db7  -
```

Docker rejects to pull this image:

```
$ docker pull evil-registry.rudy.family/busybox@sha256:deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef
evil-registry.rudy.family/busybox@sha256:deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef: Pulling from busybox
manifest verification failed for digest sha256:deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef
```

Pulling with `oci-distribution` does not fail, and returns the bogus hash to the user:

```rust
let arg = "evil-registry.rudy.family/busybox@sha256:deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef"
let image: Reference = arg.parse().unwrap();
let client = Client::default();
let auth = &oci_distribution::secrets::RegistryAuth::Anonymous;
let accepted_media_types = vec!["application/vnd.docker.image.rootfs.diff.tar.gzip"];

let (_m, d) = client.pull_manifest(&image, auth).await.unwrap();
println!("pull_manifest({}) -> {}", arg, d);

let data = client.pull(&image, auth, accepted_media_types).await.unwrap();
println!("pull({}) -> {:?}", arg, data.digest);
```

```data
pull_manifest(evil-registry.rudy.family/busybox@sha256:deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef) -> sha256:deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef
pull(evil-registry.rudy.family/busybox@sha256:deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef) -> Some("sha256:deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef")
```

The spec [says this](https://github.com/opencontainers/distribution-spec/blob/570d0262abe8ec5e59d8e3fbbd7be4bd784b200e/spec.md?plain=1#L155) regarding content verification:

> Most clients MAY ignore the [`Docker-Content-Digest` header], but if it is used, the client MUST verify the value against the uploaded blob data.

The full reproducer is available at https://github.com/burgerdev/evil-registry/.

thomastaylor312:
This implements missing functionality around using the `Docker-Content-Digest` header to validate the digest of incoming content. This should now be spec compliant.

Additionally, as a safety measure, if a digest is given as part of a image reference, that digest will also be used if the header isn't set or it will be validated if the algorithm is the same

Fixes #149


thomastaylor312:
I did test this by using and modifying @burgerdev's helpful reproducer: https://github.com/burgerdev/evil-registry/. I am honestly not sure what we want to do here for testing this as we'd have to essentially create a misbehaving server like the reproducer does to run integration tests against it. I am open to any ideas to automate the testing here

thomastaylor312:
So I just went ahead and adapted @burgerdev's go example in rust for the purposes of the test

src/client.rs
```diff
@@ -1912,31 +1954,107 @@ impl TryFrom<&ChallengeRef<'_>> for BearerChallenge {
...
+                    "Incorrect digest format on header value. Should have algorithm".to_string(),
+                ))
+            })?;
+            let (ref_algo, _) = ref_digest.split_once(':').ok_or_else(|| {
+                OciDistributionError::GenericError(Some(
+                    "Incorrect digest format on image reference. Should have algorithm".to_string(),
+                ))
+            })?;
+            if algo != ref_algo {
+                return Ok(Some(digest));
```
burgerdev, line 2012: I fear that this still leaves us open to a malicious response that does not match the ref digest.

Browsing the spec a bit more, I think it should be safe to verify both the requested and the canonical digest, because

> A GET request to an existing manifest URL MUST provide the expected manifest, with a response code that MUST be 200 OK. 
> [...]
> If the blob is not found in the registry, the response code MUST be 404 Not Found.

I read that as "the registry is NOT free to return content that does not match the request". So if the requested reference was a digest and the response code is 200, the registry found a manifest with that digest and thus the body should match it.

How about returning a `Vec` here and verifying all (0, 1 or 2) returned digests?

thomastaylor312: It might be a little weird, but I can do it. I'll get back to it on Monday

src/blob.rs
```diff
@@ -0,0 +1,162 @@
...
+    Sha512(sha2::Sha512),
+}
+
+impl Digester {
+    pub fn new(digest: &str) -> crate::errors::Result<Self> {
+        let (algo, _) = digest.split_once(':').ok_or_else(|| {
+            OciDistributionError::GenericError(Some(format!(
+                "Digest header value is not in the expected format: {}",
+                digest
+            )))
```
flavio, line 93: nitpicking: I would create a dedicated error type for that

src/blob.rs
```diff
@@ -0,0 +1,162 @@
...
+        })?;
+
+        match algo {
+            "sha256" => Ok(Digester::Sha256(sha2::Sha256::new())),
+            "sha384" => Ok(Digester::Sha384(sha2::Sha384::new())),
+            "sha512" => Ok(Digester::Sha512(sha2::Sha512::new())),
+            _ => Err(OciDistributionError::GenericError(Some(format!(
+                "Unsupported digest algorithm: {}",
+                algo
+            )))),
```
flavio, line 103: nitpicking: I would create a dedicated error type for that

src/blob.rs
```diff
@@ -0,0 +1,162 @@
...
+            Box::pin(futures_util::stream::iter(vec![Ok(Bytes::from_static(
+                data,
+            ))])),
+            Digester::new(incorrect_sha).unwrap(),
+            incorrect_sha.to_string(),
+        );
+        stream
+            .try_collect::<Vec<_>>()
+            .await
+            .expect_err("Should error with invalid sha");
```
flavio, line 160: By using a dedicated error type to identify the invalid digest format, we could write a more specific assert. We could make sure the returned error is of the type we expected (`using `matches!`)

src/client.rs
```diff
@@ -1912,31 +1954,107 @@ impl TryFrom<&ChallengeRef<'_>> for BearerChallenge {
...
+fn digest_header_value(headers: HeaderMap) -> Result<Option<String>> {
+    headers
+        .get(DOCKER_DIGEST_HEADER)
+        .map(|hv| hv.to_str().map(|s| s.to_string()))
+        .transpose()
+        .map_err(|e| {
+            OciDistributionError::GenericError(Some(format!(
+                "Cannot convert Docker-Content-Digest header value to string: {}",
+                e
+            )))
```
flavio, line 1967: nitpick: another type of error?

src/client.rs
```diff
@@ -1912,31 +1954,107 @@ impl TryFrom<&ChallengeRef<'_>> for BearerChallenge {
...
+        })
+}
+
+/// Given the optional digest header value and digest of the reference, returns the digest of the
+/// content, validating that the digest of the content matches the proper digest
+fn validate_digest(
```
flavio, line 1973: Dropping the comment here, but it applies to the other functions introduced by this PR. I feel `client.rs` is growing too much, I would like to split that into smaller modules to improve maintenance.
While this should be done with a future PR, maybe these functions should already be placed inside of an internal module.

src/client.rs
```diff
@@ -1912,31 +1954,107 @@ impl TryFrom<&ChallengeRef<'_>> for BearerChallenge {
...
+        (Some(digest), None) => Ok(Some(digest)),
+        (None, Some(digest)) => Ok(Some(digest.to_string())),
+        (Some(digest), Some(ref_digest)) => {
+            // According to the spec, it is just fine if the requested digest differs from the
+            // canonical digest from the headers if the algorithms are different. If they are the
+            // same, then the digests should match.
+            let (algo, _) = digest.split_once(':').ok_or_else(|| {
+                OciDistributionError::GenericError(Some(
+                    "Incorrect digest format on header value. Should have algorithm".to_string(),
+                ))
```
flavio, line 2004: Sorry for being a pain, but I would like to have dedicated error types. Otherwise it feels like going back to the early days when everything was just a `anyhow::Error`

src/client.rs
```diff
@@ -1912,31 +1954,107 @@ impl TryFrom<&ChallengeRef<'_>> for BearerChallenge {
...
+            })?;
+            let (ref_algo, _) = ref_digest.split_once(':').ok_or_else(|| {
+                OciDistributionError::GenericError(Some(
+                    "Incorrect digest format on image reference. Should have algorithm".to_string(),
+                ))
```
flavio, line 2009: Same here

src/client.rs
```diff
@@ -1912,31 +1954,107 @@ impl TryFrom<&ChallengeRef<'_>> for BearerChallenge {
...
         }
-        Some(hv) => hv
-            .to_str()
-            .map(|s| s.to_string())
-            .map_err(|e| OciDistributionError::GenericError(Some(e.to_string()))),
     }
 }
 
+/// Helper for calculating and validating the digest of the given content
+fn calculate_and_validate(content: &[u8], digest: &str) -> Result<String> {
```
flavio, line 2028: It would be nice to have unit tests for these helper functions that have been introduced

src/errors.rs
```diff
@@ -99,11 +101,35 @@ pub enum OciDistributionError {
     #[error("Failed to convert Config into ConfigFile: {0}")]
     /// Transparent wrapper around `std::string::FromUtf8Error`
     ConfigConversionError(String),
+    /// Verification of the content digest failed
+    #[error("Failed to verify content digest. {0}")]
+    ContentDigestVerificationError(ContentDigestVerificationError),
```
flavio, line 106: Why going this way instead of doing something like:

```rust
#[error("Failed to verify content digest. Expected {expected:?}, got {actual:?}")]
ContentDigestVerificationError{
  expected: String,
  actual: String,
},
```

The other changes to this file can then be dropped.

tests/digest_validation.rs
```diff
@@ -0,0 +1,279 @@
+use std::net::SocketAddr;
+
```
flavio, line 2: nitpicking, the empty line can be removed. Also, some comment at the top of the file explaining what is going on would be nice

tests/digest_validation.rs
```diff
@@ -0,0 +1,279 @@
...
+    let reference = Reference::try_from(format!(
+        "{}/busybox@sha256:deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef",
+        server.server
+    ))
+    .expect("failed to parse reference");
+
+    client
+        .pull_manifest(&reference, auth)
+        .await
+        .expect_err("Expected an error with a mismatched sha");
```
flavio, line 146: With typed errors we could write an assertion about the specific error type we get back

flavio (changes requested):
Thanks for having looked into this issue. I left some comments

src/client.rs
thomastaylor312: Yeah this is why I created the new `blob.rs` file. I can break these out too

src/client.rs
thomastaylor312: I did think about adding these errors, but the error enum is quite large. What if I added a new `digest.rs` file and had a separate enum for digest errors that would be included in the top level error type?

src/errors.rs
thomastaylor312: I actually had it this way first, then thought I was going to need to reuse the struct elsewhere and didn't end up using it. I'll put it back and/or refactor when I handle the other digest errors

tests/digest_validation.rs
thomastaylor312: Actually the line should be there. The common pattern for separating imports in order is 

- core
- std
- external crates
- internal crate imports

Otherwise they're just all smooshed together.

I honestly wish this was a cargo fmt thing but it isn't. I do know that when rust-analyzer adds imports it separates them as well last time I dug into it

src/client.rs
```diff
@@ -751,19 +739,35 @@ impl Client {
...
         } else {
             let status = res.status();
             let headers = res.headers().clone();
+            trace!(headers = ?res.headers(), "Got Headers");
             let body = res.bytes().await?;
             validate_registry_response(status, &body, &url)?;
 
-            digest_header_value(headers, None)
+            // SAFETY: We just checked that the header exists, so this is safe to unwrap
+            let header_digest = digest_header_value(headers)?.unwrap();
```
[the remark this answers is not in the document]
thomastaylor312: I left it this way for a reason, but now that I am looking at it again I can't remember why. I'll do it this way and then comment if I remember why I was keeping it the same way

src/client.rs
flavio: That would be nice!

tests/digest_validation.rs
flavio: Thanks for the explanation, I didn't realize that.

thomastaylor312:
@flavio @burgerdev I think I have addressed all comments and this should be ready for a re-review

flavio (approved):
Thanks for the changes!

thomastaylor312:
@burgerdev going to leave this open until tomorrow in case you want to do a re-review

src/client.rs
```diff
@@ -748,8 +735,29 @@ impl Client {
...
             .await?;
 
-        trace!(headers=?res.headers(), "Got Headers");
-        if res.headers().get("Docker-Content-Digest").is_none() {
+        if let Some(digest) = digest_header_value(res.headers().clone())? {
+            let status = res.status();
+            let body = res.bytes().await?;
+            validate_registry_response(status, &body, &url)?;
+
+            // If the reference has a digest and the digest header has a matching algorithm, compare
```
burgerdev, line 743: If the reference has a digest and the digest header has a different algorithm, the returned digest is not verified. Is that intentional?

burgerdev (commented):
Hi @thomastaylor312,

Sorry that I'm only getting back to this now. LGTM, but I left one comment.

Cheers, Markus

src/client.rs
thomastaylor312: Yes, because the behavior here is unclear. We are only `HEAD`-ing the manifest here and returning the digest in the header if it exists. Perhaps the proper behavior of this function is to only ever fetch the complete manifest, which we could definitely change in the next version

burgerdev: Makes sense, and there is probably no good use case for fetching the digest of a manifest by digest :)

PR #156, opened 2024-08-09
chore(*): Bumps crate version and some dependencies

thomastaylor312:
I also took care of updating our cargo deny action and configuration


flavio (approved):
Thanks!

PR #160, opened 2024-08-19
client: add support to set https_proxy/no_proxy for client

Xynnn007:
This feature allows a client to pull images via a proxy. In some cases like intranet with a outgoing proxy it's very useful.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -12,7 +12,7 @@ use http::{HeaderValue, StatusCode};
 use http_auth::{parser::ChallengeParser, ChallengeRef};
 use olpc_cjson::CanonicalFormatter;
 use reqwest::header::HeaderMap;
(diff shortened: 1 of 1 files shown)


thomastaylor312 (approved):
Thanks for adding this. I think having this around is probably a good idea. Would also like to note that `reqwest` does try to respect the `HTTPS_PROXY` and `NO_PROXY` env vars if they are set. But I think allowing this to be set programmatically is a good idea

Xynnn007:
> Thanks for adding this. I think having this around is probably a good idea. Would also like to note that `reqwest` does try to respect the `HTTPS_PROXY` and `NO_PROXY` env vars if they are set. But I think allowing this to be set programmatically is a good idea

Thanks. As we have multiple reqwest client inside a same process, env would influence all clients which is not expected.

PR #163, opened 2024-09-08
Allow doing start-end RANGE requests against blobs

thundergolfer:
Addresses https://github.com/oras-project/rust-oci-client/issues/162

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1051,7 +1051,7 @@ impl Client {
         layer: impl AsLayerDescriptor,
         mut out: T,
     ) -> Result<()> {
-        let response = self.pull_blob_response(image, &layer, None).await?;
+        let response = self.pull_blob_response(image, &layer, None, None).await?;
 
         let mut maybe_header_digester = digest_header_value(response.headers().clone())?
             .map(|digest| Digester::new(&digest).map(|d| (d, digest)))
@@ -1130,7 +1130,11 @@ impl Client {
         image: &Reference,
         layer: impl AsLayerDescriptor,
     ) -> Result<SizedStream> {
-        stream_from_response(self.pull_blob_response(image, &layer, None).await?, layer)
+        stream_from_response(
+            self.pull_blob_response(image, &layer, None, None).await?,
+            layer,
+            true,
+        )
     }
 
     /// Stream a single layer from an OCI registry starting with a byte offset.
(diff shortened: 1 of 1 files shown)


flavio (commented):
Seems fine, can you address the build failure?

Moreover, is there some way we can test this feature inside of our CI?

thundergolfer:
Oh damn sorry I thought this was compiling because I was pointing at my fork but I had a cached `main` 😖

thundergolfer:
@flavio I updated the implementation and added a test. During reimplementation I switch to using `pull_blob_stream_partial` and ran into an issue where `VerifyingStream` doesn't work when you get a partial response back (the digests will never match). 

I'm not sure how this was working before? Let me know what you think :)

src/client.rs
```diff
@@ -1142,14 +1146,19 @@ impl Client {
...
 
         let status = response.status();
         match status {
-            StatusCode::OK => Ok(BlobResponse::Full(stream_from_response(response, &layer)?)),
+            StatusCode::OK => Ok(BlobResponse::Full(stream_from_response(
+                response, &layer, true,
+            )?)),
             StatusCode::PARTIAL_CONTENT => Ok(BlobResponse::Partial(stream_from_response(
-                response, &layer,
+                response, &layer, false,
```
thundergolfer, line 1161: If the `offset` was ever non-zero this method should already have been failing on stream verification? I checked registry spec and couldn't see anything about returned specialized digests for partial reads.

flavio:
@thundergolfer thanks for having introduced the test case. 

@thomastaylor312 can you take a quick look too?

flavio (approved):

thundergolfer:
Nice, thanks. Let me know if you need anything else before merge 👌

src/client.rs
thomastaylor312: It makes sense that partial pulls can't be validated, so we should probably document the `pull_blob_stream_partial` that partial pulls by nature cannot be validated and as such should be validated by the caller when complete

thomastaylor312 (changes requested):

thomastaylor312:
Sorry forgot to add context when I hit the submit button. This all looks good to me. One minor point around adding some clarification to the docs for a function around the security implications

src/client.rs
thomastaylor312: Actually, now that I'm looking at this more, we might need to return the two possible digests to validate against to the caller on a partial request, otherwise they won't have access to the digest of the header (probably as new data members of `SizedStream`). What do other clients do/expect to handle digest validation in these cases?

thomastaylor312: @thundergolfer If you'd like/are willing to do this, that would be great! If not, let me know and I'll switch this to an approval and I'll change it after merge since it is a preexisting edge case we didn't account for originally

thundergolfer: Hey @thomastaylor312, I've switched focus off this project so I'll be unlikely to get to it soon.

thomastaylor312: Ok sounds good. I'll merge this as is and fix up. Thanks for adding this!

thomastaylor312 (approved):

PR #166, opened 2024-09-25
Docs.rs badge, update crate name.

bddap:
Looks like the crate has been re-published as `oci-client`. This pr updates the docs link in the readme.


bddap:
Btw @thomastaylor312 , you seem to be the only owner on that crate https://crates.io/crates/oci-client

PR #177, opened 2024-10-13
feat: Add option for using monolithic push

joonas:
Following up on #175 and #176, this rounds out the remaining set of functionality that's needed to use `rust-oci-client` for pushing to GCP Artifact Registry similarly to how `oras-go` can be used for this.

This seemed like an option that should be available for configuring the client in case you're working with a registry that's not fully spec compliant.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -559,6 +559,10 @@ impl Client {
         data: &[u8],
         digest: &str,
     ) -> Result<String> {
+        if self.config.use_monolithic_push {
+            return self.push_blob_monolithically(image_ref, data, digest).await;
+        }
+
         match self.push_blob_chunked(image_ref, data, digest).await {
             Ok(url) => Ok(url),
             Err(OciDistributionError::SpecViolationError(violation)) => {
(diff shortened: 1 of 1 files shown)


src/client.rs
```diff
@@ -559,14 +559,18 @@ impl Client {
...
     ) -> Result<String> {
-        match self.push_blob_chunked(image_ref, data, digest).await {
-            Ok(url) => Ok(url),
-            Err(OciDistributionError::SpecViolationError(violation)) => {
-                warn!(?violation, "Registry is not respecting the OCI Distribution Specification when doing chunked push operations");
-                warn!("Attempting monolithic push");
-                self.push_blob_monolithically(image_ref, data, digest).await
+        if self.config.use_monolithic_push {
+            self.push_blob_monolithically(image_ref, data, digest).await
+        } else {
```
flavio, line 564: Just a matter of personal taste, I would rewrite it get rid of the extra indentation caused by the `else` statement:

```rust
if self.config.use_monolithic_push {
  return self.push_blob_monolithically(image_ref, data, digest).await;
} 
```

flavio (changes requested):
Thanks for the change, I left a minor comment

src/client.rs
joonas: Great suggestion, I've adjusted the code accordingly 👍

thomastaylor312 (approved):
Good to go whenever you're ok with it @flavio

flavio (approved):
thanks for having done the requested change

PR #178, opened 2024-10-15
Add pull referrers to Client

rylev:
Adds an API for fetching the referrers manifest list for a given reference. 

This implementation will error if passed a reference with a tag instead of a digest. The OCI server I'm using for testing (harbor) errors when requesting against a tag and not a digest. I'm unsure if we should be trying to resolve that tag to a digest on the client first, so for now I've let it error.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1486,6 +1486,32 @@ impl Client {
         ret
     }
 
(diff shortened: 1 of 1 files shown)


thomastaylor312 (approved):
Thanks @rylev! I did double check the spec and it specifically says digest, so I think a tag should error

PR #181, opened 2024-10-24
Add subject to OciImageManifest

fibonacci1729:
This allows creating accessory artifacts that can be associated with artifacts in a repository. Additionally added a missing error code.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -191,6 +191,8 @@ pub enum OciErrorCode {
     NameInvalid,
     /// Repository name is not known
     NameUnknown,
+    /// Manifest is not found
+    NotFound,
     /// Provided length did not match content length
     SizeInvalid,
     /// Manifest tag did not match URI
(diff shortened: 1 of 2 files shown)


bacongobbler:
Out of curiosity, where did you find the NOT_FOUND error code? I cannot find it listed under https://github.com/opencontainers/distribution-spec/blob/main/spec.md#error-codes

There is a mention of [legacy docker support error codes](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#legacy-docker-support-error-codes) which states "these error codes are OPTIONAL and clients SHOULD NOT depend on them". I did notice both `TagInvalid` and `ManifestUnverified` is listed under `OciErrorCode` though, so maybe we already set precedent for these codes?

fibonacci1729:
I was seeing a serde json deserialization error saying "not found" was an unexpected error code variant when querying for a manifest that didn't exist. Adding this variant fixed it and allowed me to check the code to determine existence of an artifact.

fibonacci1729:
@thomastaylor312 I think this is ready for review if you get a chance

bacongobbler:
@fibonacci1729 just for reference this was with Harbor, correct?

https://github.com/goharbor/harbor

PR #185, opened 2024-12-10
chore: Updates thiserror and fixes cargo deny

thomastaylor312:
This updates our deny.toml to remove unused exceptions and change which unicode license is being allowed. I also fixed a small clippy error that showed up


flavio (approved):
thanks, I was stuck with other tasks and I didn't have the time to look into the issue

PR #189, opened 2025-02-20
feat: Add blocking client

dave-tucker:
Uses the reqwuest/blocking feature to add a client that can be used from sync rust. This is gated behind the `blocking` feature.

Removed test-registry from default-features.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,8 +8,6 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
     steps:
       - uses: actions/checkout@v4.2.2
       - uses: engineerd/configurator@v0.0.10
--- a/src/blocking.rs
+++ b/src/blocking.rs
@@ -1,23 +1,19 @@
 //! OCI distribution client for fetching oci images from an OCI compliant remote store
 use std::collections::HashMap;
 use std::convert::TryFrom;
-use std::sync::Arc;
+use std::io::Write;
 use std::time::Duration;
 
-use futures_util::future;
-use futures_util::stream::{self, BoxStream, StreamExt, TryStreamExt};
 use http::header::RANGE;
-use http::{HeaderValue, StatusCode};
+use http::HeaderValue;
 use http_auth::{parser::ChallengeParser, ChallengeRef};
 use olpc_cjson::CanonicalFormatter;
+use reqwest::blocking::{RequestBuilder, Response};
 use reqwest::header::HeaderMap;
-use reqwest::{NoProxy, Proxy, RequestBuilder, Response, Url};
+use reqwest::{NoProxy, Proxy, Url};
 use serde::Serialize;
-use tokio::io::{AsyncWrite, AsyncWriteExt};
-use tokio::sync::RwLock;
 use tracing::{debug, trace, warn};
 
-pub use crate::blob::*;
 pub use crate::types::*;
 
 use crate::digest::{digest_header_value, validate_digest, Digest, Digester};
@@ -30,7 +26,7 @@ use crate::manifest::{
 use crate::secrets::RegistryAuth;
 use crate::secrets::*;
 use crate::sha256_digest;
-use crate::token_cache::{RegistryOperation, RegistryToken, RegistryTokenType, TokenCache};
+use crate::token_cache::{RegistryOperation, RegistryToken, RegistryTokenType, SyncTokenCache};
 use crate::Reference;
 
 const MIME_TYPES_DISTRIBUTION_MANIFEST: &[&str] = &[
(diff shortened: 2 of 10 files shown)


src/blocking.rs
```diff
@@ -0,0 +1,2830 @@
...
+/// Default value for `ClientConfig::max_concurrent_upload`
+pub const DEFAULT_MAX_CONCURRENT_UPLOAD: usize = 16;
+
+/// Default value for `ClientConfig::max_concurrent_download`
+pub const DEFAULT_MAX_CONCURRENT_DOWNLOAD: usize = 16;
+
+/// Default value for `ClientConfig:default_token_expiration_secs`
+pub const DEFAULT_TOKEN_EXPIRATION_SECS: usize = 60;
+
+static DEFAULT_USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"));
```
thomastaylor312, line 50: Let's move this out of here and in `client.rs` to `lib.rs` so it isn't duplicated across both blocking and the normal client

src/token_cache.rs
```diff
@@ -179,3 +180,114 @@ impl TokenCache {
...
+#[derive(Clone)]
+pub(crate) struct SyncTokenCache {
+    // (registry, repository, scope) -> (token, expiration)
+    tokens: BTreeMap<TokenCacheKey, TokenCacheValue>,
+    /// Default token expiration in seconds, to use when claim doesn't specify a value
+    pub default_expiration_secs: usize,
+}
+
+#[cfg(feature = "blocking")]
+impl SyncTokenCache {
```
thomastaylor312, line 194: Can you take the logic in these functions that is the same and break them out to functions that can be reused rather than copy pasting?

src/blocking.rs
```diff
@@ -0,0 +1,2830 @@
...
+                .map(|at| format!("?artifactType={at}"))
+                .unwrap_or_default(),
+        ))
+    }
+}
+
+/// The OCI spec technically does not allow any codes but 200, 500, 401, and 404.
+/// Obviously, HTTP servers are going to send other codes. This tries to catch the
+/// obvious ones (200, 4XX, 5XX). Anything else is just treated as an error.
+fn validate_registry_response(status: reqwest::StatusCode, body: &[u8], url: &str) -> Result<()> {
```
thomastaylor312, line 1307: This function looks like it can be shared between both clients

src/blocking.rs
```diff
@@ -0,0 +1,2830 @@
...
+            let message = res.text()?;
+            Err(OciDistributionError::ServerError { url, code, message })
+        }
+    }
+
+    /// Helper function to convert location header to URL
+    ///
+    /// Location may be absolute (containing the protocol and/or hostname), or relative (containing just the URL path)
+    /// Returns a properly formatted absolute URL
+    fn location_header_to_url(
```
thomastaylor312, line 1211: This method could probably be moved to the `ClientConfig` struct as a `pub(crate)` so it can be reused across the clients

src/blocking.rs
```diff
@@ -0,0 +1,2830 @@
...
+                "{scheme}://{registry}{lh}",
+                scheme = self.config.protocol.scheme_for(registry)
+            ))
+        } else {
+            Ok(lh.to_string())
+        }
+    }
+
+    /// Convert a Reference to a v2 manifest URL.
+    fn to_v2_manifest_url(&self, reference: &Reference) -> String {
```
thomastaylor312, line 1229: Same note here around moving this to a method on `ClientConfig` for better sharability. Basically all the rest of the methods until line 1302 can probably be moved over there for better reuse

src/blocking.rs
```diff
@@ -0,0 +1,2830 @@
...
+pub struct Certificate {
+    /// Which encoding is used by the certificate
+    pub encoding: CertificateEncoding,
+
+    /// Actual certificate
+    pub data: Vec<u8>,
+}
+
+/// A client configuration
+pub struct ClientConfig {
```
thomastaylor312, line 1460: We should do whatever we need to here to make sure this is shared across both blocking and non-blocking clients. Pretty much everything here until the tests can be shared across clients

src/blocking.rs
```diff
@@ -0,0 +1,2830 @@
...
+
+        Ok(BearerChallenge {
+            realm: realm.into_boxed_str(),
+            service,
+        })
+    }
+}
+
+#[cfg(test)]
+mod test {
```
thomastaylor312, line 1720: I would really prefer if we could find a way to test both clients using the same tests. Even if we just had an enum with the blocking and non blocking client (that we only used for tests) that matches on the client and calls the right thing

thomastaylor312 (changes requested):
I am all onboard with having a blocking client as well! The main focus of the comments here are trying to reuse as much code as possible between the two clients. Some parts will just be easier or need to be copy/pasted and modified, but I think a good chunk of them could be shared, which will reduce maintenance burden/problems

.github/workflows/build.yml
```diff
@@ -8,8 +8,6 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    strategy:
```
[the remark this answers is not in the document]
dave-tucker: Yes. strategy only applies to matrix jobs. From the [docs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast)
> jobs.<job_id>.strategy.fail-fast applies to the entire matrix.

dave-tucker:
@thomastaylor312 thanks for the review. I'll address your comments and try and get another revision done with more re-use and less copy/paste some time in the next couple of weeks 😄

PR #191, opened 2025-02-24
feat: Direct Bearer Token Authentication

prabhpreet:
Adds RegistryAuth::Bearer for use of direct bearer token authentication


thomastaylor312 (approved):
Thanks for adding this! Just noting for our purposes that when we release, this is a breaking change since it is a new enum member

PR #194, opened 2025-03-20
feat: add HTTP proxy for client

Xynnn007:
This commit adds one more configuration for client about proxy, s.t. http_proxy. http_proxy sets the http proxy that the client uses to pull images, while no_proxy will define the remote address that will not be accessed by proxy.

This is useful in a local registry scenario where original pull request is in http.

Typical https_proxy looks like `http://127.0.0.1:5432` no_proxy looks like `127.0.0.1,192.168.1.1`

Note that `client_builder.proxy()` function will add rather than overwrite the proxies to underlying reqwest::Client, thus no side effections upon the https proxy setting.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -328,6 +328,15 @@ impl TryFrom<ClientConfig> for Client {
             client_builder = client_builder.proxy(proxy);
         }
 
+        if let Some(proxy_addr) = &config.http_proxy {
+            let no_proxy = config
+                .no_proxy
+                .as_ref()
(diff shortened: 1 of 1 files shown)


flavio (approved):

flavio:
let's wait for the other maintainers to review the change, then we can merge it

thomastaylor312 (approved):

thomastaylor312:
Good on my side here. @flavio feel free to merge when ready

Xynnn007:
kindly ping. Do we have any release schedule for this?

PR #204, opened 2025-06-09
add push blob async stream support

lambinoo:
Good morning/afternoon!

This is an attempt at upstream some of the changes I've been carrying for this crate!

This adds a `push_blob_stream` method to the client that takes a stream as input. It's pretty much the same code as the chunked push method, except it just takes data until there's none left. Its signature is such that one can just use the stream from the streaming pull blob method.

It also adds a method that verifies if a blob exists. Combined with the streaming push method, it's quite handy to easily copy a bunch of blobs between two repositories.

Let me know if you see a need to split this into two pull requests or anything else.

I'll add an example and maybe a test if it's required!

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -635,13 +635,10 @@ impl Client {
     ) -> Result<String> {
         let mut location = self.begin_push_chunked_session(image).await?;
         let mut start: usize = 0;
-        loop {
-            (location, start) = self
-                .push_chunk(&location, image, &blob_data[start..], start)
-                .await?;
-            if start >= blob_data.len() {
-                break;
-            }
+
+        for chunk in blob_data.chunks(self.push_chunk_size) {
+            let chunk = bytes::Bytes::from(chunk.to_vec());
+            (location, start) = self.push_chunk(&location, image, chunk, start).await?;
         }
         self.end_push_chunked_session(&location, image, blob_digest)
             .await
@@ -660,8 +657,9 @@ impl Client {
         let mut range_start = 0;
 
         while let Some(blob_data) = blob_data_stream.next().await {
-            // Make sure we respect the max chunk size
-            for chunk in blob_data?.chunks(self.push_chunk_size) {
+            let mut blob_data = blob_data?;
+            while !blob_data.is_empty() {
+                let chunk = blob_data.split_to(self.push_chunk_size.min(blob_data.len()));
                 (location, range_start) = self
                     .push_chunk(&location, image, chunk, range_start)
                     .await?;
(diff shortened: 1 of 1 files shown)


thomastaylor312 (approved):
This is great! Sorry for the delay in reviewing. Accidentally marked my notification as read.

thomastaylor312:
@lambinoo Could you please do a simple amend of your commit and force push? For some reason the tests didn't kick off

flavio (changes requested):
The code looks good to me, but it would better to have some unit tests covering that.

@lambinoo: our tests spin up a real registry and then do operation against it. It would be great to test these new functions against a real registry

lambinoo:
I added some unit tests (good thing we did, it just didn't work.. it just happened that the use case we have it for now is only pushing data smaller than the max chunk size..). Each feature should have it in its own commit to keep things clear!

Let me know if you see anything else

flavio (changes requested):
Thanks for the tests! I've left some minor comments, once these changes are in we can merge the PR

lambinoo:
Fixed up! Crossing fingers it's all good now!

src/client.rs
```diff
@@ -1360,35 +1413,38 @@ impl Client {
...
-            start_byte + self.push_chunk_size - 1
-        } else {
-            blob_data.len() - 1
-        };
-        let body = blob_data[start_byte..end_byte + 1].to_vec();
+
+        let range_size = self.push_chunk_size.min(blob_data.len());
+        let end_range_inclusive = range_start + range_size - 1;
+
+        let body = blob_data[..range_size].to_vec();
```
thomastaylor312, line 1431: I don't know how I missed this the first time, but this shouldn't use `to_vec()`. This means you'd end up cloning each chunk of data as it is pushed. The http request body should be able to just take the slice made here

thomastaylor312 (commented):
Sorry @lambinoo, there is one thing I missed in my last review that should be fixed. Everything else looks good

thomastaylor312 (changes requested):
Sorry forgot to request changes

src/client.rs
flavio: We have to provide something that is compatible with [`RequestBuilder.body`](https://docs.rs/reqwest/0.12.22/reqwest/struct.RequestBuilder.html#method.body).

The [`Body`](https://docs.rs/reqwest/0.12.22/reqwest/struct.Body.html) struct has the `From` trait implemented for: `&'static [u8]`, `&'static str`, `Bytes`, `File`, `String` and `Vec<u8>`.

I think in our case, the only viable solution would be to move to [`Bytes`](https://docs.rs/bytes/1.10.1/bytes/struct.Bytes.html)

lambinoo: I believe the last commit should fix that up. I described the changes in the commit message

flavio (approved):

thomastaylor312 (approved):
Yeah, I still don't love the allocations here, but as has been demonstrated, that is outside of the scope of this PR, so this should be good to go! Thanks again @lambinoo

PR #209, opened 2025-08-02
refactor(client)!: Reduces allocations of streaming data

thomastaylor312:
This is a breaking change, but modifies the API so anything that would be chunks of data we send (so blobs and the config data) is now a `Bytes`. That means any cloning of data is in the user's hands as they control how they construct the `Bytes`. We only clone bytes, which is cheap and doesn't clone the underlying data. The one exception that I didn't change to `Bytes` is `pull_manifest_raw` that returns a `Vec<u8>` since that is often handled in memory anyway. For consistency, we can change that to return `Bytes` as well if we'd like

Also includes some clippy fixes that will start erroring in the latest version of rust

This is followup to discussion we had in #204


thomastaylor312:
Sweet, I've swapped that out as well.

PR #210, opened 2025-08-09
chore: replace jwt with jsonwebtoken

BradLugo:
https://github.com/oras-project/rust-oci-client/issues/167 suggests moving away from `jwt` because it's not actively maintained. These changes replace `jwt` with `jsonwebtoken`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ chrono = { version = "0.4.23", features = ["serde"] }
 futures-util = "0.3"
 http = "1.1"
 http-auth = { version = "0.1", default-features = false }
-jwt = "0.16"
+jsonwebtoken = "9.3"
 lazy_static = "1.4"
 oci-spec = "0.8.1"
 olpc-cjson = "0.1"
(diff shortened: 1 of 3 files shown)


BradLugo:
Hi @thomastaylor312! Yeah, I haven't had the time yet, but I want to write up some thoughts about replacing the `jwt`. Now that I know why we need to interact with JWTs and how, I'm not so sure this change is particularly useful. Quick summary of what I've been thinking:
- `rust-oci-client` doesn't use the `jwt` crate for much, and, as far as RustSec is concerned, it's not marked as unmaintained (yet)
- The `jsonwebtoken` crate doesn't currently have an ergonomic way to retrieve the token's expiration
- Should `rust-oci-client` be interacting with JWTs in the way that it currently is in the first place? For example, maybe this crate should only acknowledge the information provided by the registry's API response and remove looking at the JWT to see if there's an expiration field there (which in turn would remove any dependency on JWT-related crates).

thomastaylor312 (approved):
@BradLugo I think you make some good points. We might be able to get rid of JWT deps altogether. But I do think swapping out to a crate that is more well-maintained is good for now, even if it isn't ergonomic. 

@flavio before I merge, did you have any thoughts?

PR #246, opened 2026-03-12
feat: add catalog endpoint for listing repositories

michael-herwig:
## Summary
- Add `catalog()` method implementing the OCI Distribution Spec `/v2/_catalog` endpoint
- Add `CatalogResponse` struct for typed deserialization
- Add `to_catalog_url()` URL helper following the `to_list_tags_url()` pattern
- Support pagination via `n` (page size) and `last` (last repo from previous page)
- Add unit test for URL generation and integration test with testcontainers

## Details

The implementation follows the existing `list_tags()` pattern exactly:
- URL helper for constructing the endpoint URL
- `store_auth_if_needed` → build request with optional query params → `RequestBuilderWrapper` → `apply_auth` → send → validate → deserialize
- Returns `CatalogResponse` (consistent with `TagResponse` pattern) rather than `Vec<String>`

## Test plan
- [x] Unit test `test_to_catalog_url` verifies URL generation for standard and mirror registries
- [x] Integration test `test_catalog` (gated behind `test-registry` feature) pushes to two repos, verifies both appear in catalog response, and tests pagination with `n`/`last` parameters

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -88,6 +88,13 @@ pub struct TagResponse {
     pub tags: Vec<String>,
 }
 
+/// The data returned by a successful catalog request.
+#[derive(Deserialize, Debug)]
+pub struct CatalogResponse {
+    /// List of available repositories in the registry.
+    pub repositories: Vec<String>,
+}
+
 /// Layer descriptor required to pull a layer
 pub struct LayerDescriptor<'a> {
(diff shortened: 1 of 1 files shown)


flavio (approved):
Thanks for your contribution and sorry for the delay

PR #253, opened 2026-03-31
feat: Add Node.js bindings via napi-rs

davidfestal:
Introduces Node.js bindings for the `rust-oci-client` using `napi-rs`, enabling Node.js applications to interact with OCI registries.

#### Features:
- 19 OciClient methods: pull, push, manifests (pull/push/raw), blobs (pull/push/exists/mount), tags, catalog, referrers, platform-aware manifest resolution
- Authentication helpers: anonymous, basic, bearer
- Platform resolver configuration via os/arch/variant filter
- TLS configuration: custom CA certificates, exclusive certificate roots, accept-invalid-certificates
- Proxy support (HTTP/HTTPS/no-proxy)
- OCI annotation and media type constants
- Full TypeScript definitions auto-generated

#### Supported platforms (9 targets):
- macOS x86_64 and ARM64
- Linux x86_64 and ARM64 (glibc)
- Linux x86_64 and ARM64 (musl/Alpine)
- Linux s390x and ppc64le (glibc)
- Windows x86_64

#### CI/CD:
- Build and test workflows for all platforms
- Publish workflow for NPM releases
- Node.js 22 and 24 support
- Integration tests with Zot registry (push/pull round-trips)
- s390x tested via QEMU system VM with interpreter-only V8 flags
- TLS tests with self-signed certificates

#### Prerequisites for publishing:

The package is scoped as `@oras-project/oci-client`. Before publishing to npmjs, the following is needed:

1. **npm organization:** An `oras-project` [npm organization](https://docs.npmjs.com/creating-an-organization) must be created to own the `@oras-project` scope.

2. **Authentication — two options:**

   - **Option A — Trusted Publishers (recommended):** npm now supports [Trusted Publishing](https://docs.npmjs.com/trusted-publishers) via OIDC, eliminating the need for long-lived secrets. The GitHub Actions workflow is already configured with `id-token: write` permissions. A trusted publisher would need to be configured on npmjs.com for this package, linking the `nodejs-publish-call.yml` workflow. This provides automatic provenance attestations and short-lived, workflow-scoped credentials.

   - **Option B — NPM_TOKEN secret:** A classic automation token can be created on npmjs.com and stored as an `NPM_TOKEN` repository secret. This is simpler to set up but requires manual token rotation and carries the usual risks of long-lived credentials.

Assisted-by: Cursor


davidfestal:
cc @TerryHowe This is the work I briefly told you about last week at KubeCon EU when discussing at the ORAS booth.

davidfestal:
Before opening this PR, the result of the bindings publish workflow has already been pushed from my fork to my personal scope on NPMJS: [@dfatwork/oci-client](https://www.npmjs.com/package/@dfatwork/oci-client).
It can be easily tested from there in any Node.js project.

davidfestal:
cc @sabre1041

TerryHowe:
> cc @TerryHowe This is the work I briefly told you about last week at KubeCon EU when discussing at the ORAS booth.

Ah, okay. Thanks. I'm not a maintainer on this project though, so they will need to chime in.

.github/workflows/release.yml
```diff
@@ -3,11 +3,26 @@ on:
   push:
     tags:
       - "v*"
+
 jobs:
   publish:
     name: publish to crates.io
     runs-on: ubuntu-latest
+    needs: build-and-test-nodejs
```
TerryHowe: **Critical:** This couples the Rust crate publish to the Node.js bindings build. If the napi-rs build fails for any reason (cross-compilation toolchain issue, napi version bump, etc.), the Rust crate cannot be released even if the Rust code itself is fine. These should be independent — the `publish` job should not `needs: build-and-test-nodejs`. Node.js bindings can have their own separate release trigger.

bindings/nodejs/src/lib.rs
```diff
@@ -0,0 +1,1181 @@
...
+        }
+
+        if let Some(no_proxy) = &self.no_proxy {
+            config.no_proxy = Some(no_proxy.clone());
+        }
+
+        if let Some(p) = &self.platform {
+            let os = Os::Other(p.os.clone());
+            let arch = Arch::Other(p.architecture.clone());
+            let variant = p.variant.clone();
```
TerryHowe: **Critical:** `Os::Other("linux")` will not compare equal to `Os::Linux`, and `Arch::Other("amd64")` will not compare equal to `Arch::Amd64`. The `oci-spec` crate deserializes known values like `"linux"` and `"amd64"` into their typed enum variants, not `Other(...)`. This means the platform filter will silently never match a standard multi-platform image.

Use the `FromStr` impl (or `From<&str>` if available) to parse into the correct variant:
```rust
let os: Os = p.os.parse().unwrap_or_else(|_| Os::Other(p.os.clone()));
let arch: Arch = p.architecture.parse().unwrap_or_else(|_| Arch::Other(p.architecture.clone()));
```

Cargo.toml
```diff
@@ -1,3 +1,16 @@
+[workspace]
+members = [".", "bindings/nodejs"]
```
TerryHowe, line 2: **Important:** Adding `bindings/nodejs` to the workspace means `cargo build`, `cargo test`, and `cargo check` run by existing Rust contributors will now attempt to build the napi crate, which requires the napi-build tooling. This may break contributor workflows that don't have Node.js/napi set up, and will slow down plain Rust builds.

Consider whether the bindings should be a separate workspace (not part of the root workspace), or document clearly in the README/CONTRIBUTING that `napi-build` is now a prerequisite for workspace-level builds.

.github/workflows/nodejs-build-call.yml
```diff
@@ -0,0 +1,332 @@
...
+  MACOSX_DEPLOYMENT_TARGET: '10.13'
+  CARGO_INCREMENTAL: '1'
+  CARGO_TERM_COLOR: always
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
```
TerryHowe: **Important:** Third-party actions should be pinned to a full commit SHA to prevent supply-chain attacks (a tag like `@v6` can be moved). This is inconsistent with the existing `release.yml` which already pins `actions/checkout` to a SHA. All third-party `uses:` entries in this workflow (and `nodejs-publish-call.yml`) should be pinned, e.g.:
```yaml
uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v6
```
Tools like [Ratchet](https://github.com/sethvargo/ratchet) or `pinact` can automate this.

.github/workflows/nodejs-publish-call.yml
```diff
@@ -0,0 +1,76 @@
...
+    secrets:
+      NPM_TOKEN:
+        required: true
+  workflow_dispatch:
+
+jobs:
+  publish:
+    name: Publish to npm
+    runs-on: ubuntu-latest
+    needs: build-and-test-nodejs
```
TerryHowe, line 14: **Important:** This causes the Node.js build+test to run twice during a release: once from `release.yml` (which calls this file via `publish-nodejs-bindings`, which in turn declares `needs: build-and-test-nodejs` here) and once directly from `release.yml`'s own `build-and-test-nodejs` job. The binaries are rebuilt and re-uploaded rather than reused. Consider having `release.yml` pass the already-built artifacts through, or remove the redundant `needs` here and rely solely on `release.yml` for sequencing.

TerryHowe (changes requested):
Great work overall — the binding coverage, test setup, and multi-platform CI are well done. Inline comments cover the critical and important issues. The two blockers are the `release.yml` coupling (Rust publish shouldn't depend on Node.js build success) and the platform filter bug (`Os::Other`/`Arch::Other` won't match typed oci-spec variants for standard platforms like linux/amd64).

bindings/nodejs/src/lib.rs
davidfestal: Fixed

.github/workflows/nodejs-publish-call.yml
```diff
@@ -0,0 +1,76 @@
...
+    permissions:
+      contents: write
+      id-token: write
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Setup node
+        uses: actions/setup-node@v6
+        with:
```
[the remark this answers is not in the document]
davidfestal: Fixed

.github/workflows/nodejs-build-call.yml
davidfestal: Fixed

Cargo.toml
davidfestal: Fixed by adding the `default-members` line.

.github/workflows/nodejs-build-call.yml
```diff
@@ -0,0 +1,332 @@
...
+
+      - name: Setup node
+        uses: actions/setup-node@v6
+        with:
```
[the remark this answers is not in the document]
davidfestal: Fixed

bindings/nodejs/src/lib.rs
```diff
@@ -0,0 +1,1181 @@
...
+            config.platform_resolver = Some(Box::new(move |manifests| {
```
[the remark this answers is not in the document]
davidfestal: Fixed

bindings/nodejs/README.md
```diff
@@ -0,0 +1,212 @@
...
+console.log(`Pulled ${imageData.layers.length} layers`);
+console.log(`Digest: ${imageData.digest}`);
+
+// Push an image
+const response = await client.push(
+  'registry.example.com/myimage:v1',
+  layers,
+  config,
+  basicAuth('username', 'password'),
+  null // Let the client generate the manifest
```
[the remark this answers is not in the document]
davidfestal: The generated TypeScript type for this parameter is `manifest?: ImageManifest | undefined | null` because `napi-rs` maps `Rust Option<T>` to `T | undefined | null`, so passing `null` is valid even with `strictNullChecks`.

bindings/nodejs/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+
+# OCI spec types for Platform arch/os enums
+oci-spec.workspace = true
+
+# Async runtime
+tokio = { version = "1", features = ["full"] }
+
+# Bytes for efficient buffer handling
+bytes = "1"
+
```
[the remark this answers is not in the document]
davidfestal: Fixed

bindings/nodejs/package.json
```diff
@@ -0,0 +1,81 @@
+{
+  "name": "@oras-project/oci-client",
+  "version": "0.16.1",
+  "description": "Node.js bindings for rust-oci-client - OCI Distribution client",
+  "main": "index.js",
+  "types": "index.d.ts",
+  "files": [
+    "index.d.ts",
+    "index.js"
```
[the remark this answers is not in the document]
davidfestal: review comment is false.

.github/workflows/nodejs-build-call.yml
```diff
@@ -0,0 +1,332 @@
...
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            ~/.napi-rs
+            .cargo-cache
+            target/
+          key: ${{ matrix.settings.target }}-cargo-${{ matrix.settings.host }}
```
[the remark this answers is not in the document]
davidfestal: Fixed

.github/workflows/release.yml
davidfestal: Fixed

.github/workflows/nodejs-publish-call.yml
davidfestal: Fixed

.github/workflows/release.yml
```diff
@@ -3,11 +3,26 @@ on:
...
+  # Build and test Node.js bindings FIRST
+  # This also ensures version sync before any publishing
+  build-and-test-nodejs:
+    name: Build & Test Node.js
+    uses: ./.github/workflows/nodejs-build-call.yml
+
+  # Publish Node.js packages to npm
+  publish-nodejs-bindings:
+    name: Publish to npm
+    uses: ./.github/workflows/nodejs-publish-call.yml
```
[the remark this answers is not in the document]
davidfestal: Fixed

davidfestal:
@TerryHowe I fixed all your comments.

davidfestal:
@thomastaylor312 @bacongobbler @flavio any feedback on this PR ?

davidfestal:
@TerryHowe @thomastaylor312 @bacongobbler @flavio

I rebased this PR on top of `main` (after `0.17.0` bump), updated it to accommodate some API changes, completed it with missing APIs and platforms, and fixed some remaining limitations.
But your workflow approval is needed to run the pending checks, and your review to move forward.

The published NPM package (rescoped to my personal NPM org), is available as [@dfatwork/oci-client](https://www.npmjs.com/package/@dfatwork/oci-client), published from my fork by this workflow run : https://github.com/davidfestal/rust-oci-client/actions/runs/28392305994/job/84122982207

Thanks for your attention.

cc @sabre1041

.github/workflows/nodejs-build-call.yml
flavio, line 1: Can you please point to a specific version of the action (major.minor.patch)?

Also, I don't *think* we need to use the ratcheck tool. Dependabot is already taking care of updating the shasum and the commented human-readable version on its own.

flavio (commented):
Thanks for all the work that went into this PR. However, I have some concerns about merging it.

I don't have good enough knowledge of JavaScript/TypeScript to maintain this codebase and keep the bindings working and up-to-date across all the releases. The PR also adds a significant amount of CI (the npm build matrix covers 12+ targets) that would become part of this repo's maintenance burden.

Wouldn't it be possible to move the bindings to a dedicated project? Looking at `bindings/nodejs/src/lib.rs`, the bindings only consume the public API of the crate, so the path dependency could simply become a regular crates.io dependency. This is also the common pattern for language bindings, which usually live in their own repo with their own release cadence.

CC @thomastaylor312

thomastaylor312:
After looking at this, I agree with @flavio. I think it would be really nice to have, but I'd prefer if it lives _separately_ from this repo. It would definitely be a nice way to have an OCI client for node, but we don't have the experience here to maintain it as part of this repo. We do try and document any breaking API changes, so you would know what is coming when you want to update

flavio:
After having discussed with @thomastaylor312, we both think it's better to close this PR.

@davidfestal: please, let us know if there's anything we need to change inside of the crate to make your job easier.

davidfestal:
> After having discussed with @thomastaylor312, we both think it's better to close this PR.
> 
> @davidfestal: please, let us know if there's anything we need to change inside of the crate to make your job easier.

Thanks !
I'll come back to you with any feedback as soon as I have updated my code to make the javascript bindings a dedicated project.

PR #259, opened 2026-05-15
fix implement referrers fallback

flavio:
The distribution/distribution container registry does not implement the native OCI referrers API (/v2/.../referrers/<digest>), returning 404 for those requests.

The OCI Distribution Spec requires clients to fall back to the referrers tag schema when the native API returns 404, see:

- [oci-referrers](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers)
- [oci-tag-schema](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#referrers-tag-schema)

Kinda unrelated, but while building docs locally to verify the new inline docs I ran into some doc warnings. This PR addresses them.

BTW: this is required to be able to pull Sigstore bundle signatures from OCI registries, since they are pushed as referrers.

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,6 +1,9 @@
 build +FLAGS='':
     cargo build {{FLAGS}}
 
+doc:
+    RUSTDOCFLAGS="--cfg docsrs -D warnings" cargo +nightly doc --workspace --all-features --no-deps
+
 test:
     cargo fmt --all -- --check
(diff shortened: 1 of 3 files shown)


thomastaylor312 (approved):

flavio:
Thanks, I'll fix the cargo deny on another PR

PR #260, opened 2026-05-21
fix: do not dump sensitive data in debug mode

### LINKED ISSUE #254: auth token getting logged at debug level
i noticed the auth token is [dumped in clear text](https://github.com/oras-project/rust-oci-client/blob/v0.16.1/src/client.rs#L801) on auth successful at debug level. i don't see need to do this. if the team is onboard, i can file a fix.

flavio:
Do not print registry's auth token and registry password when running in debug mode.

Fixes https://github.com/oras-project/rust-oci-client/issues/254

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -805,7 +805,7 @@ impl Client {
         match auth_res.status() {
             reqwest::StatusCode::OK => {
                 let text = auth_res.text().await?;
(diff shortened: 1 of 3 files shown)


flavio:
@thomastaylor312: I would like to cut a patch release on this gets merged

thomastaylor312 (approved):

PR #265, opened 2026-05-29
fix: URL-encode artifactType in pull_referrers query

### LINKED ISSUE #264: `pull_referrers` filter silently returns empty: `artifactType` query value isn't URL-encoded
### Summary

`Client::pull_referrers(..., Some(artifact_type))` returns an empty index whenever the artifact type contains a `+` (i.e. every IANA `+json` / `+xml` / `+yaml` media type — SBOMs, signatures, attestations, the lot). The value is interpolated into the query string verbatim; standard query-string semantics decode `+` as space, so `application/spdx+json` reaches the registry as `application/spdx json` and matches nothing.

### Source

[`src/client.rs`](https://github.com/oras-project/rust-oci-client/blob/main/src/client.rs) — `Client::to_v2_referrers_url`:

```rust
at = artifact_type
    .map(|at| format!("?artifactType={at}"))
    .unwrap_or_default(),
```

No percent-encoding of the value. Same in 0.14 and 0.17.

### Reproduction (curl only, no Rust required)

Start a fresh OCI 1.1 registry:

```bash
docker run -d --rm -p 5099:5000 ghcr.io/project-zot/zot-linux-amd64:latest
```

Run the script in the appendix below to push a subject manifest with two referrers (SPDX + CycloneDX SBOMs). Then:

```bash
SUBJECT=<digest emitted by the setup script>
BASE=http://localhost:5099/v2/demo/referrers-bug/referrers/$SUBJECT

# 1. Unfiltered — both referrers, artifactType populated
$ curl -s "$BASE"
{"schemaVersion":2,"mediaType":"application/vnd.oci.image.index.v1+json","manifests":[
  {"...","digest":"sha256:a75ef…","artifactType":"application/spdx+json"},
  {"...","digest":"sha256:8efb0…","artifactType":"application/vnd.cyclonedx+json"}
]}

# 2. Filter with raw `+` (what oci-client sends) — EMPTY
$ curl -s "$BASE?artifactType=application/spdx+json"
{"schemaVersion":2,"mediaType":"application/vnd.oci.image.index.v1+json","manifests":[]}

# 3. Same filter with `+` encoded as `%2B` — works
$ curl -s "$BASE?artifactType=application/spdx%2Bjson"
{"schemaVersion":2,"mediaType":"application/vnd.oci.image.index.v1+json","manifests":[
  {"...","digest":"sha256:a75ef…","artifactType":"application/spdx+json"}
]}
```

Verified against zot v2.1.17 (`distSpecVersion: 1.1.1`). Any registry that follows standard `application/x-www-form-urlencoded` query-string decoding has the same behaviour — that's most of them.

### Rust-side reproduction

Same registry, same data:

```rust
let referrers = client
    .pull_referrers(&subject_ref, Some("application/spdx+json"))
    .await?;
assert!(!referrers.manifests.is_empty()); // fails — empty
```

### Why it matters

The `+suffix` convention covers essentially every modern OCI referrer payload type:

- `application/spdx+json`, `application/vnd.cyclonedx+json` — SBOMs
- `application/vnd.dev.cosign.artifact.sig.v1+json` — Cosign signatures
- `application/vnd.in-toto+json` — attestations

So the `?artifactType=` filter is effectively broken for the common case. Today's workaround is either to filter client-side (cheap now that `ImageIndexEntry::artifact_type` is populated as of 0.17) or to bypass the helper and build the URL manually.

### Fix

Percent-encode the value before interpolating. Minimal change in `to_v2_referrers_url`:

```diff
-    at = artifact_type
-        .map(|at| format!("?artifactType={at}"))
-        .unwrap_or_default(),
+    at = artifact_type
+        .map(|at| format!(
+            "?artifactType={}",
+            percent_encoding::utf8_percent_encode(at, percent_encoding::NON_ALPHANUMERIC),
+        ))
+        .unwrap_or_default(),
```

`percent-encoding` is already in the transitive tree via `reqwest`. Happy to send the PR.

### Appendix: setup script

<details>
<summary>setup.sh — pushes one subject + two SBOM referrers</summary>

```bash
#!/usr/bin/env bash
set -euo pipefail
BASE=http://localhost:5099/v2/demo/referrers-bug

push_blob() {                                                            # $1=content
  local digest="sha256:$(printf '%s' "$1" | sha256sum | awk '{print $1}')"
  local loc=$(curl -sX POST -D - "$BASE/blobs/uploads/" -o /dev/null \
    | grep -i '^location:' | tr -d '\r\n' | sed 's/.*: //')
  local sep='?'; echo "$loc" | grep -q '?' && sep='&'
  curl -sX PUT "http://localhost:5099${loc}${sep}digest=$digest" \
    -H 'Content-Type: application/octet-stream' --data-binary "$1" > /dev/null
  echo "$digest"
}
push_manifest() {                                                        # $1=body $2=ref
  curl -sX PUT "$BASE/manifests/$2" \
    -H 'Content-Type: application/vnd.oci.image.manifest.v1+json' \
    --data-binary "$1" -D /dev/stderr -o /dev/null 2>&1 \
    | grep -iE '^(HTTP|Oci-Subject):'
}

EMPTY=$(push_blob '{}')
SUBJECT_BLOB=$(push_blob 'subject-blob')
SPDX_BLOB=$(push_blob '{"spdxVersion":"SPDX-2.3"}')
CDX_BLOB=$(push_blob '{"bomFormat":"CycloneDX","specVersion":"1.5"}')

SUBJECT='{"schemaVersion":2,"mediaType":"application/vnd.oci.image.manifest.v1+json","artifactType":"application/vnd.demo.subject.v1+json","config":{"mediaType":"application/vnd.oci.empty.v1+json","digest":"'$EMPTY'","size":2},"layers":[{"mediaType":"application/vnd.demo.subject.blob.v1","digest":"'$SUBJECT_BLOB'","size":12}]}'
SUBJECT_DIGEST="sha256:$(printf '%s' "$SUBJECT" | sha256sum | awk '{print $1}')"
SUBJECT_SIZE=$(printf '%s' "$SUBJECT" | wc -c)
push_manifest "$SUBJECT" "$SUBJECT_DIGEST"

for tuple in "spdx+json:$SPDX_BLOB:25" "vnd.cyclonedx+json:$CDX_BLOB:45"; do
  AT="application/${tuple%%:*}"; rest="${tuple#*:}"; BLOB="${rest%:*}"; SIZE="${rest##*:}"
  M='{"schemaVersion":2,"mediaType":"application/vnd.oci.image.manifest.v1+json","artifactType":"'$AT'","config":{"mediaType":"application/vnd.oci.empty.v1+json","digest":"'$EMPTY'","size":2},"layers":[{"mediaType":"'$AT'","digest":"'$BLOB'","size":'$SIZE'}],"subject":{"mediaType":"application/vnd.oci.image.manifest.v1+json","digest":"'$SUBJECT_DIGEST'","size":'$SUBJECT_SIZE',"artifactType":"application/vnd.demo.subject.v1+json"}}'
  D="sha256:$(printf '%s' "$M" | sha256sum | awk '{print $1}')"
  push_manifest "$M" "$D"
done

echo "subject digest: $SUBJECT_DIGEST"
```

</details>

---

*Assisted by: Claude Code (Opus 4.8)*

obassett:
Fixes #264

`Client::to_v2_referrers_url` interpolated the `artifactType` filter into the
query string verbatim. Standard query-string decoding treats `+` as a space, so
`+json`/`+xml`/`+yaml` media types (SBOMs, signatures, attestations) reached the
registry mangled and the filter matched nothing — see #264 for full context and
reproduction.

This percent-encodes the value with `NON_ALPHANUMERIC` before interpolating, and
promotes `percent-encoding` to a direct dependency (already present transitively
via `reqwest`). The tag-schema fallback is unaffected — it filters client-side by
direct string comparison, not via the URL.

Adds `test_to_v2_referrers_url` covering the `+json` encoding case and the
no-filter case. `cargo test`, `cargo fmt --check`, and
`cargo clippy --all-targets -- -D warnings` all pass locally.

---

*Assisted by: Claude Code (Opus 4.8)*

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,6 @@ jsonwebtoken = { version = "10.3", default-features = false }
 lazy_static = "1.4"
 oci-spec = "0.9"
 olpc-cjson = "0.1"
-percent-encoding = "2.3"
 regex = "1.11"
 reqwest = { version = "0.13", default-features = false, features = [
   "json",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,7 @@ jsonwebtoken = { version = "10.3", default-features = false }
 lazy_static = "1.4"
 oci-spec = "0.9"
 olpc-cjson = "0.1"
+percent-encoding = "2.3"
 regex = "1.11"
 reqwest = { version = "0.13", default-features = false, features = [
   "json",
--- a/src/client.rs
+++ b/src/client.rs
@@ -1986,36 +1986,39 @@ impl Client {
         )
     }
 
-    /// Convert a Reference to a v2 manifest URL.
+    /// Convert a Reference to a v2 referrers URL.
     fn to_v2_referrers_url(
         &self,
         reference: &Reference,
         artifact_type: Option<&str>,
     ) -> Result<String> {
+        let digest = reference.digest().ok_or_else(|| {
+            OciDistributionError::GenericError(Some(
+                "Getting referrers for a tag is not supported".into(),
+            ))
+        })?;
+
         let registry = reference.resolve_registry();
-        Ok(format!(
-            "{scheme}://{registry}/v2/{repository}/referrers/{reference}{at}",
+        let base = format!(
+            "{scheme}://{registry}",
             scheme = self.config.protocol.scheme_for(registry),
-            repository = reference.repository(),
-            reference = if let Some(digest) = reference.digest() {
-                digest
-            } else {
-                return Err(OciDistributionError::GenericError(Some(
-                    "Getting referrers for a tag is not supported".into(),
-                )));
-            },
-            at = artifact_type
-                .map(|at| {
-                    format!(
-                        "?artifactType={}",
-                        percent_encoding::utf8_percent_encode(
-                            at,
-                            percent_encoding::NON_ALPHANUMERIC
-                        )
-                    )
-                })
-                .unwrap_or_default(),
-        ))
+        );
+        let mut url =
+            Url::parse(&base).map_err(|e| OciDistributionError::UrlParseError(e.to_string()))?;
+        url.path_segments_mut()
+            .map_err(|_| {
+                OciDistributionError::GenericError(Some(
+                    "cannot build referrers URL: base URL is cannot-be-a-base".into(),
+                ))
+            })?
+            .push("v2")
+            .extend(reference.repository().split('/'))
+            .push("referrers")
+            .push(digest);
+        if let Some(at) = artifact_type {
+            url.query_pairs_mut().append_pair("artifactType", at);
+        }
+        Ok(url.into())
     }
 }
 
--- a/src/client.rs
+++ b/src/client.rs
@@ -2005,7 +2005,15 @@ impl Client {
                 )));
             },
             at = artifact_type
-                .map(|at| format!("?artifactType={at}"))
+                .map(|at| {
+                    format!(
+                        "?artifactType={}",
+                        percent_encoding::utf8_percent_encode(
+                            at,
+                            percent_encoding::NON_ALPHANUMERIC
+                        )
+                    )
+                })
                 .unwrap_or_default(),
         ))
     }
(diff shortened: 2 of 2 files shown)


flavio:
@obassett thanks for the bug report and for the fix. I've changed your approach to rely on `url` crate to build the final url. I think is more idiomatic and bulletproof

flavio:
merging, the windows failure happened because of rate limiting on docker.io

PR #266, opened 2026-06-02
test: add regression coverage for pull_blob on a 404 (#64)

### LINKED ISSUE #64: [Bug] `Client::pull_blob` returns `Ok(...)` even when the blob wasn't found
### Steps to reproduce

The following test reproduces the issue (you can paste it under `client::test` and run it with `cargo test --features test-registry test_pull_blob_not_found`):

```rust
#[tokio::test]
#[cfg(feature = "test-registry")]
async fn test_pull_blob_not_found() {
    // initialize the registry
    let docker = clients::Cli::default();
    let test_container = docker.run(registry_image());
    let port = test_container.get_host_port_ipv4(5000);

    let c = Client::new(ClientConfig {
        protocol: ClientProtocol::HttpsExcept(vec![format!("localhost:{}", port)]),
        ..Default::default()
    });

    // Layer is not found at `image-repository`
    let image_reference: Reference = format!("localhost:{}/image-repository", port).parse().unwrap();
    let mut buf = Vec::new();
    let dummy_digest = sha256_digest(&[]);
    let pull_result = c.pull_blob(&image_reference, &dummy_digest, &mut buf).await;

    // The result is OK, but the response body does not correspond to the contents of the blob
    assert!(pull_result.is_ok());
    assert!(String::from_utf8(buf).unwrap().starts_with("{\"errors\":[{\"code\":\"BLOB_UNKNOWN\""));
}
```

### Expected result

As a user of the library, I intuitively expect `pull_blob` to return `Ok` only when the blob was found, and `Err` otherwise.

obassett:
## What

Adds an offline regression test verifying that the blob-pull methods return `Err` — and write nothing to the caller's output buffer — when the registry responds with a `404`.

## Why

Issue #64 reported that `Client::pull_blob` returned `Ok(())` even when the registry replied `404` with a `{"errors":[{"code":"BLOB_UNKNOWN",...}]}` envelope, silently writing that error body into the caller's output buffer.

The underlying bug was already fixed by commit 22c2504 ("fix: exposed errors when pulling layers"), which added an `error_for_status()?` guard before the response body is streamed. However, no test exercised the not-found path for blobs — the existing tests only covered the happy path — so the fix could silently regress.

## Changes

- `tests/digest_validation.rs`:
  - The mock registry's `404` now carries a realistic `BLOB_UNKNOWN` OCI error envelope body (previously a bare status with an empty body), so the test can detect the original body-leak failure mode.
  - New `test_pull_blob_not_found` drives a missing blob through all three entry points and asserts:
    - `pull_blob` returns `Err` **and** leaves the output buffer empty,
    - `pull_blob_stream` returns `Err`,
    - `pull_blob_stream_partial` returns `Err`.

No library code changed — this is test-only coverage that locks in existing behaviour.

## Verification

- `cargo test` — full suite passes (73 tests across lib, integration, and doc-tests; 1 pre-existing ignored).
- Confirmed the test guards the original bug: temporarily removing `error_for_status()?` at `src/client.rs:1262` makes `test_pull_blob_not_found` fail on the `buf.is_empty()` assertion ("got 73 bytes"); restoring it passes.
- `cargo clippy --tests` and `cargo fmt --check` clean.

Closes #64

---

Assisted by: Claude Code (Opus 4.8)

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1259,7 +1259,13 @@ impl Client {
         let layer_digest = layer.as_layer_descriptor().digest.to_string();
         let mut layer_digester = Digester::new(&layer_digest)?;
 
-        let mut stream = response.error_for_status()?.bytes_stream();
+        let status = response.status();
+        let url = response.url().to_string();
+        if !status.is_success() {
+            let body = response.bytes().await?;
+            return validate_registry_response(status, &body, &url);
+        }
+        let mut stream = response.bytes_stream();
 
         let mut out = pin!(out);
 
@@ -1338,6 +1344,7 @@ impl Client {
             layer,
             true,
         )
+        .await
     }
 
     /// Stream a single layer from an OCI registry starting with a byte offset. This can be used to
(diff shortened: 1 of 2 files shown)


flavio (approved):

flavio:
Thanks @obassett for your contribution, I just made some small tweaks on top of it and then I merged it.

PR #270, opened 2026-06-05
feat(token_cache): allow programmatic access to the client token cache

vdice:
The goal here is to allow programmatic access to a client's token cache, specifically/only to `insert` a token into the cache for use with subsequent registry authentication.  An applicable scenario is for proxied OCI registries wherein a bearer token valid for a generic deployment environment can also be used to authenticate with the proxy layer for OCI push/pull operations, etc.  Hence, when a user has an auth token valid both for the deployment environment and registry proxy, this token can be added to this OCI client's cache and used for OCI operations.

The current approach to achieve the goal here in this PR is to update visibility of the token cache and allow calling `insert`.  Definitely open to adjusting approach if there is a preferred alternate that can still support scenarios above.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -277,7 +277,8 @@ pub struct Client {
     config: Arc<ClientConfig>,
     // Registry -> RegistryAuth
     auth_store: Arc<RwLock<HashMap<String, RegistryAuth>>>,
-    tokens: TokenCache,
+    /// Token cache for the client
+    pub tokens: TokenCache,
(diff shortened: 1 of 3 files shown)


bacongobbler (approved):
LGTM. The only methods `TokenCache` exposes are `insert`, `get`, and `parse_expiration_from_jwt`, so this seems safe to expose.

vdice:
Thanks @bacongobbler; merge when ready.

PR #272, opened 2026-06-16
push: avoid hashing each layer twice

vadika:
## What

`Client::push` hashes every layer's content **twice**:

1. once in `OciImageManifest::build` (`manifest.rs`), which computes
   `sha256_digest(&layer.data)` for each descriptor, and
2. again in the upload loop in `push` via `ImageLayer::sha256_digest()`
   (`sha256_digest` is not memoized — it re-hashes the whole `data` each call).

For a large layer that's a full redundant SHA-256 pass over the data.

This PR reuses the digests already present on the manifest descriptors (same
order as `layers`). If a caller supplies a manifest whose layer count doesn't
match `layers`, it falls back to per-layer hashing, so behaviour is unchanged.

## Why / measurements

Pushing a **1 GiB** blob to a local `registry:2` over loopback,
`use_monolithic_push: true`, 3 runs each (AMD Threadripper PRO 5955WX, SHA-NI):

| | throughput | push time |
|---|---:|---:|
| main (double hash) | ~163 MiB/s | 5.51 s |
| this PR (single hash) | **~332 MiB/s** | **3.08 s** |

**+90%**, ~2.43 s saved (one 1 GiB SHA-256 pass). The patched high-level
`Client::push` now matches the `oras` CLI (~336–341 MiB/s) on the same setup.

## Notes

- The remaining `layer.data.clone()` in the loop is a single copy forced by the
  by-reference `&[ImageLayer]` signature (the reqwest body needs owned/`'static`
  data); it is not redundant and is left as-is.
- Unit tests (`cargo test --lib`) pass; `cargo clippy --lib` clean.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -578,15 +578,27 @@ impl Client {
             None => OciImageManifest::build(layers, &config, None),
         };
 
-        // Upload layers
-        stream::iter(layers)
-            .map(|layer| {
+        // Upload layers.
+        //
+        // Reuse the per-layer digests already computed while building (or
+        // supplied with) the manifest, rather than hashing every layer a
+        // second time here. For large layers this avoids a full redundant
+        // SHA-256 pass over the data. When `build` produced the manifest its
+        // `layers` are in the same order as `layers`; if a caller supplied a
+        // manifest whose layer count does not match, fall back to hashing each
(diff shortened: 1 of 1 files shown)


flavio (approved):
nice improvement, thanks!

PR #273, opened 2026-06-18
feat(monostream): Monolithic stream push

slakkala:
Add a Client method push_blob_stream_monolithically() that does a monolithic push-then-put instead of sending the stream chunked in PATCHes.

In tests, the chunked push was roughly 50% slower for 500 Mbit/s link, and 100% slower for 1 Gbit/s link, making it much slower than the `oras-cli` tool.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -672,33 +672,39 @@ impl Client {
             .await
     }
 
-    /// Pushes a blob to the registry as a series of chunks from an input stream as monolithic
+    /// Pushes a blob to the registry from an input stream.
     ///
-    /// Returns the pullable location of the blob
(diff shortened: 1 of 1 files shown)


flavio:
@slakkala thanks for the contribution, I've made some small changes to your PR:

- unified stream blob push into push_blob_stream
- extended test suite
- avoid usage of `unwrap`

I'll merge the PR once the CI is done and green

flavio (approved):

slakkala:
>     * unified stream blob push into push_blob_stream

This was my initial design, but was hesitant to add the `Send + 'static` to the existing func.

>     * avoid usage of `unwrap`

Nice, those were pretty much just blind copy-pasta from the other monolithic push path.

PR #279, opened 2026-06-30
feat: implement chunked blob push without upfront digest calculation

### LINKED ISSUE #278: When pushing a blob stream, compute digest on demand
The current implementation of  `push_blob_stream ` requires specifying the digest up front. For streams that process bytes on demand, this digest can only be fully computed once the stream has been consumed. Therefore, I'd like to use something like this:

```rust
pub async fn push_blob_stream<T: Stream<Item = Result<bytes::Bytes>> + Send + 'static>(
        &self,
        image: &Reference,
        blob_data_stream: T,
        // blob_digest: &str, -- REMOVE
    ) -> Result<String> {
        let mut location = self.begin_push_chunked_session(image).await?;
        let mut range_start = 0;

        let mut blob_data_stream = pin!(blob_data_stream);

        while let Some(blob_data) = blob_data_stream.next().await {
            // compute hash
            let mut blob_data = blob_data?;
            while !blob_data.is_empty() {
                let chunk = blob_data.split_to(self.push_chunk_size.min(blob_data.len()));
                (location, range_start) = self
                    .push_chunk(&location, image, chunk, range_start)
                    .await?;
            }
        }
        let blob_digest = format!(..., hash) // NEW, compute digest on demand
        self.end_push_chunked_session(&location, image, blob_digest)
            .await
    }
``` 

My use case:

Download large files from a remote HTTP location, put each file into a separate oci layer and push a container image to registry. Since the files are large and are downloaded anyway, this would ideally skip local caching and connect the streams directly.

lehmanju:
Fixes #278 

Add another method that automatically computes the blob digest while it is streamed. This allows streaming a blob directly from another source without the need to completely buffer it locally.

There is probably some room for future refactoring, the old push_blob_stream method could be changed to only do monolithic pushing since now the logic is duplicated.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -78,6 +78,16 @@ pub struct PushResponse {
     pub manifest_url: String,
 }
 
+/// The data returned by [`Client::push_blob_stream_chunked`].
+pub struct PushBlobStreamChunkedResponse {
+    /// Pullable url for the uploaded blob.
+    pub blob_url: String,
+    /// Computed digest of the uploaded blob.
+    pub blob_digest: String,
+    /// Total uploaded blob size in bytes.
+    pub size: usize,
+}
+
 /// The data returned by a successful tags/list Request
 #[derive(Deserialize, Debug)]
 pub struct TagResponse {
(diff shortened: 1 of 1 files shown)


lehmanju:
actually, the computed digest should probably returned along with the upload url. that way it can be used when building a manifest

lehmanju:
OK, fixed it. I guess its ready for review, Ive used the API a lot in my local testing and didn't feel the need to change anything else.

lehmanju:
@flavio hey, did you have any chance to look at this yet?

flavio:
superseded by https://github.com/oras-project/rust-oci-client/pull/292

PR #282, opened 2026-07-15
feat: Enabled native-tls feature to use newly introduced json-webtoken-openssl as a cryptoprovider dependency.

### LINKED ISSUE #319: Align fully with Fedora crypto policies
https://github.com/trusted-execution-clusters/operator/blob/df17b032172001d6ec86c5f73af4a864851a47aa/tests/no_disallowed_crypto.rs#L27-L28

SpaceFace02:
native-tls feature should ideally use the platform specific tls instead of `rust-crypto`. For linux, its openssl. `jsonwebtoken-openssl` is a new crate which exposes this functionality for the linux platform.

This PR introduces the change to change the dependency of native-tls to this new crate.

In the production build, rust-oci-client uses jsonwebtoken's `dangerous::insecure_decode` method to parse expiration date from the jwt, which doesn't rely on a cryptobackend, but the dev dependencies include real crypto provider tests for verifying and encoding header, claims and key to a token.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ maintenance = { status = "actively-developed" }
 
 [features]
 default = ["rustls-tls", "test-registry"]
-native-tls = ["reqwest/native-tls", "jsonwebtoken/rust_crypto"]
+native-tls = ["reqwest/native-tls", "dep:jsonwebtoken-openssl"]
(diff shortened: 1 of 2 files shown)


SpaceFace02:
Hi, any update on this?

flavio:
superseded by https://github.com/oras-project/rust-oci-client/pull/286

Issue #132, opened 2024-05-22
validate_registry_response function


IIEIIEJI:
Hello!.
Maybe, in the **_validate_registry_response_** function, there is a non-compliance with the https://github.com/opencontainers/distribution-spec/blob/main/spec.md#errors-2 standard. 
Firstly the returned format may not be json ( ...A 4XX response code from the registry MAY return a body in any format....). Secondly some registries return their own code, for example Harbor v2.9.0 returns this error: r#"{"errors":[{"code": "NOT_FOUND", "message": "artifact image:0.0.1 not found"}]}]}"#.
It may be more correct to use the following variant of the **_validate_registry_response_** function: 
```
fn validate_registry_response(status: reqwest::StatusCode, body: &[u8], url: &str) -> Result<()> {
    debug!(STATUS=?status);
    match status {
        reqwest::StatusCode::OK => Ok(()),
        reqwest::StatusCode::UNAUTHORIZED => Err(OciDistributionError::UnauthorizedError {
            url: url.to_string(),
        }),
        s if s.is_success() => Err(OciDistributionError::SpecViolationError(format!(
            "Expected HTTP Status {}, got {} instead",
            reqwest::StatusCode::OK,
            status,
        ))),
        s if s.is_client_error() => {
            let text = std::str::from_utf8(body)?;
            
            // According to the OCI spec, we should see an error in the message body.
            match serde_json::from_str::<OciEnvelope>(text) {
                Ok(envelope) => Err(OciDistributionError::RegistryError {
                    envelope,
                    url: url.to_string(),
                }),
                Err(err) => {
                    debug!(err=?err,"Not valid OciError (https://github.com/opencontainers/distribution-spec/blob/master/spec.md#errors-2)");
                    Err(OciDistributionError::ServerError {
                        code: s.as_u16(),
                        url: url.to_string(),
                        message: text.to_string(),
                    })
                }
            }
        }
        s => {
            let text = std::str::from_utf8(body)?;

            Err(OciDistributionError::ServerError {
                code: s.as_u16(),
                url: url.to_string(),
                message: text.to_string(),
            })
        }
    }
}
```

thomastaylor312:
Looks like you are correct here! Code also looks pretty good with the only suggestion being to check the response headers to see if content-type is set before parsing JSON. Completely open to a PR if you'd like to submit!

renshao:
I believe this issue can be closed, with #290 now merged. cc @thomastaylor312

PR #290, opened 2026-09-12
fix: handle unknown error codes and non UTF-8 server error bodies

renshao:
Follow-up to #132. Most of that issue was already addressed (the 4xx fallback to `ServerError` in 5cf221f and `OciErrorCode::NotFound` in 3a7131d). This PR covers the two remaining gaps in `validate_registry_response`.

## Changes

- **Unknown error codes**: add `OciErrorCode::Other(String)` (a `#[serde(untagged)]` variant). Registries that return codes outside the OCI spec now still produce a `RegistryError` with the original code preserved, instead of falling back to a plain `ServerError` and losing the envelope structure.
- **Non UTF-8 bodies on non-4xx errors**: the catch-all arm (mostly 5xx, but also 1xx, 3xx and non-standard codes) now uses `String::from_utf8_lossy`, matching the 4xx arm. Previously `std::str::from_utf8(body)?` returned a UTF-8 decoding error and the HTTP status code was lost.
- Bump the minimum `serde` to `1.0.181`, which is required for untagged variants in a tagged enum.

## Compatibility

Adding a variant to the public `OciErrorCode` enum is a breaking change for exhaustive `match`es without a wildcard arm.

## Testing

- `errors::test::test_deserialize_unknown_code`: unknown code parses and round-trips through serialization.
- `client::test::test_validate_registry_response_unknown_error_code`: 4xx with an unknown code yields `RegistryError`.
- `client::test::test_validate_registry_response_server_error_non_utf8_body`: 5xx with a non UTF-8 body yields `ServerError` with the status code.

Note: `cargo clippy --all-targets -- -D warnings` fails on current `main` with recent clippy (`useless_borrows_in_formatting` on `format!("Bearer {}", &token)` in the test module). That's unrelated and left untouched here.


Cargo.toml
```diff
@@ -49,7 +49,7 @@ reqwest = { version = "0.13", default-features = false, features = [
   "stream",
 ] }
 serde_json = "1.0"
-serde = { version = "1.0", features = ["derive"] }
+serde = { version = "1.0.181", features = ["derive"] }
```
flavio: do we really need to pick this specific patch level?

renshao: Good catch, it isn't needed. Per-variant `#[serde(untagged)]` landed in serde 1.0.164, and dependencies like serde_json already require ^1.0.220.

flavio (approved):

PR #297, opened 2026-09-18
ci: fix zizmor findings for dependabot and cargo-deny action

flavio:
Dependabot updates had no cooldown period. Add a 7-day cooldown to the cargo and github-actions blocks, so new releases wait before dependabot opens a pull request for them.

The comment next to the pinned cargo-deny-action commit named the wrong version. Update the comment to v2.1.1, which matches the pinned commit.

**Update:** I've noticed we were using an old action that was using a deprecated version of node. I pushed a second commit that replaces two separate actions with taiki-e/install-action.


flavio:
@thomastaylor312 I've added trusted publishing too via commit 9c30f9e0adef28cea1b30d2ff85d23501e00da84

thomastaylor312 (approved):

